PR #1, opened 2018-11-06
TLS renegotiation

arykalin:
Adding optional TLS renegotiation for vCert client
Options:
always = RenegotiateFreelyAsClient
once = RenegotiateOnceAsClient
empty or other = RenegotiateNever

See https://golang.org/pkg/crypto/tls/#RenegotiationSupport


tr1ck3r:
This update will be needed when the TPP WebSDK supports client certificate auth but we should consider pairing this update with support for client certificate auth in VCert.

mkorejo:
I have a PR incoming for client certificate auth with TLS renegotiation enabled. Need to cut some red tape first but I look forward to working with the Venafi open source team to get this merged.

PR #62, opened 2020-01-09
Master ven54271 customfields

mr-tron:


pkg/venafi/tpp/connector_test.go
```diff
@@ -356,6 +356,9 @@ func TestRequestCertificate(t *testing.T) {
 	req.Subject.Province = []string{"Nevada"}
 	req.Subject.Country = []string{"US"}
 	req.FriendlyName = cn
+	req.CustomFields = []certificate.CustomField{
+		{Name: "custom", Value: "2019-10-10"},
+	}
```
arykalin: If possible we need to check that custom field was set on certificate

arykalin:
It's not clear to me what happens when we renew certificate with custom field, we need to test it

PR #81, opened 2020-02-23
Add metadata save to tpp19.2 and lower

brantpeery:
Added metadata to the TestRequestCertificateServiceGenerated test. 

- In order for the tests to pass, the test TPP servers must have metadata fields "TestField" and "TestField2" both of type string and enabled for certificates.

- I changed the tests to include the 'configuration' OAuth scope. This is needed as the code calls to /SystemStatus/Version that needs this privilege. **It will break users who are currently using metadata in 19.3 as they will need to include this scope by default in any enroll call that includes metadata.**

I tested against TPP 19.2.1.7205 and 19.3.0.4463 which represented the edge cases for the added logic.

### DIFF
--- a/pkg/venafi/tpp/connector_test.go
+++ b/pkg/venafi/tpp/connector_test.go
@@ -24,9 +24,6 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"fmt"
-	"github.com/Venafi/vcert/pkg/certificate"
-	"github.com/Venafi/vcert/pkg/endpoint"
-	"github.com/Venafi/vcert/test"
 	"net/http"
 	"net/url"
 	"os"
@@ -34,6 +31,10 @@ import (
 	"strings"
 	"testing"
 	"time"
+
+	"github.com/Venafi/vcert/pkg/certificate"
+	"github.com/Venafi/vcert/pkg/endpoint"
+	"github.com/Venafi/vcert/test"
 )
 
 var ctx *test.Context
@@ -54,7 +55,7 @@ func init() {
 
 	resp, err := tpp.GetRefreshToken(&endpoint.Authentication{
 		User: ctx.TPPuser, Password: ctx.TPPPassword,
-		Scope: "certificate:approve,delete,discover,manage,revoke;"})
+		Scope: "certificate:approve,delete,discover,manage,revoke;configuration"})
 	if err != nil {
 		panic(err)
 	}
@@ -439,7 +440,10 @@ func TestRequestCertificateServiceGenerated(t *testing.T) {
 	req.CsrOrigin = certificate.ServiceGeneratedCSR
 	req.FetchPrivateKey = true
 	req.KeyPassword = "newPassw0rd!"
-
+	req.CustomFields = []certificate.CustomField{
+		certificate.CustomField{Type: certificate.CustomFieldPlain, Name: "TestField", Value: "Value 1"},
+		certificate.CustomField{Type: certificate.CustomFieldPlain, Name: "TestField2", Value: "TRUE"},
+	}
 	config.UpdateCertificateRequest(req)
 
 	pickupId, err := tpp.RequestCertificate(req)
--- a/pkg/venafi/tpp/connector_test.go
+++ b/pkg/venafi/tpp/connector_test.go
@@ -55,7 +55,7 @@ func init() {
 
 	resp, err := tpp.GetRefreshToken(&endpoint.Authentication{
 		User: ctx.TPPuser, Password: ctx.TPPPassword,
-		Scope: "certificate:approve,delete,discover,manage,revoke;configuration"})
+		Scope: "certificate:approve,delete,discover,manage,revoke;"})
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/venafi/tpp/connector_test.go
```diff
@@ -54,7 +55,7 @@ func init() {
 
 	resp, err := tpp.GetRefreshToken(&endpoint.Authentication{
 		User: ctx.TPPuser, Password: ctx.TPPPassword,
-		Scope: "certificate:approve,delete,discover,manage,revoke;"})
+		Scope: "certificate:approve,delete,discover,manage,revoke;configuration"})
```
tr1ck3r: @brantpeery which method(s) are you using that require the `configuration` scope?  Most of the metadata methods are scopeless and that decision was recently affirmed by the TPP architects.

Also, not introduced here but something we should fix, there is no `approve` restriction defined for the `certificate` scope.

tr1ck3r: Looks like it is `GET SystemStatus/Version`.  I'll check that with the TPP architect, too.

tr1ck3r:
@arykalin @mr-tron this update builds on the work you did to add Custom Field support for TPP versions 19.3 and higher.

pkg/venafi/tpp/connector_test.go
arykalin: @tr1ck3r @BrandonRoehl I checked in documentation, SystemStatus/Version really needs configuration scope. The problem here that if we merge this pull request any request with custom fields will need this scope.

arykalin: Maybe change RequestSystemVersion to the additional request flag, something like pre192-cutom-field? It is not so convenient, but don't need configuration scope.

pkg/venafi/tpp/connector_test.go
```diff
@@ -439,7 +440,10 @@ func TestRequestCertificateServiceGenerated(t *testing.T) {
 	req.CsrOrigin = certificate.ServiceGeneratedCSR
 	req.FetchPrivateKey = true
 	req.KeyPassword = "newPassw0rd!"
-
+	req.CustomFields = []certificate.CustomField{
+		certificate.CustomField{Type: certificate.CustomFieldPlain, Name: "TestField", Value: "Value 1"},
+		certificate.CustomField{Type: certificate.CustomFieldPlain, Name: "TestField2", Value: "TRUE"},
+	}
```
arykalin: We already test custom fields in TestRequestAndSearchCertificate

tr1ck3r:
@brantpeery I've been discussing with the team and @arykalin came up with another idea that is preferable to adding a new requirement for the `configuration` scope. Since token auth wasn't introduced until 19.2 (one release before `GET Certificates/Request` was enhanced to support Custom Fields), we'd only need to check the version using `GET SystemStatus/Version` when a token **isn't** used for authentication (or, alternatively, always use your Custom Field implementation when **not** using a token).  That would mean pre-19.3 customers who wanted to specify Custom Fields would have to authenticate using username/password.

The following logic would check whether a token has been used:
```
if len(req.CustomFields) > 0 {
	if c.accessToken == "" {
		//Check to see if this is 19.2 or lower
		tppVersion, err := c.RequestSystemVersion()
		if err != nil {
			return "", fmt.Errorf("%s", err)
		}
				.....
	}
}
```

brantpeery:
How about change the logic to 

1. Make the call that requests the certificate
2. If there are metatdata items in the request, check the returned DN for at least one metadata item (this lets us know if it is 19.3 or not)
3. If metadata items are null on the server, then use the 19.2 logic to save the metadata items.

pkg/venafi/tpp/connector_test.go
brantpeery: I will look at that test to make sure it will test this logic. If so, I will remove these tests so we don't have multiple custom field requirements

brantpeery:
I removed the dependency on SystemStatus/Version so we don't have to have additional auth scope. The only path this could take that would put it into an error state is if enroll was used like renew to update an existing cert that already had metadata fields. 
Is that situation likely to ever happen?

arykalin:
Hi @brantpeery ok, it's looks good for me. We recently merged new code into master so we need to resolve merge conflict.

arykalin:
@brantpeery could clarify the renew case you're referring to?

brantpeery:
Did the team intend to make the CustomField structure so it can only handle single value custom fields? I am going to match that structure in the compare that will determine if the system is 19.3 or not. However, I would think that multiple values would be a requirement here as TPP allows it.

```go
type CustomField struct {
	Type  CustomFieldType
	Name  string
	Value string
}
```

mr-tron:
You can use multiple customfields for this.
```
request.CustomFields = []CustomField{
     {Name: "field1": Value: "value1"},
     {Name: "field1": Value: "value2"},
     {Name: "field2": Value: "value3"},
}
```

mr-tron:
I think we should add this to documentation.

PR #85, opened 2020-02-27
Omit sans

mr-tron:


mr-tron:
all tests passed

pkg/certificate/certificate.go
```diff
@@ -276,7 +276,9 @@ func GenerateRequest(request *Request, privateKey crypto.Signer) error {
 func (request *Request) GenerateCSR() error {
 	certificateRequest := x509.CertificateRequest{}
 	certificateRequest.Subject = request.Subject
-	certificateRequest.DNSNames = request.DNSNames
+	if !request.OmitSANs {
+		certificateRequest.DNSNames = request.DNSNames
+	}
 	certificateRequest.EmailAddresses = request.EmailAddresses
 	certificateRequest.IPAddresses = request.IPAddresses
 	certificateRequest.URIs = request.URIs
```
tr1ck3r: @mr-tron do you think we should be ignoring all SAN types since the option is general `--omit-sans`?  It won't matter for the customer issue because in general public CAs only support DNS SANs but I'm thinking it is probably best for the option to apply to all SAN types we support.

mr-tron: so you suggest to move emails, IPs and URIs inside `if` condition?

pkg/venafi/tpp/connector.go
```diff
@@ -493,7 +495,32 @@ func (c *Connector) RenewCertificate(renewReq *certificate.RenewalRequest) (requ
...
+			return "", err
+		}
+		statusCode, _, _, err := c.request("PUT", urlResourceCertificate+urlResource(guid), r)
+		if err != nil {
+			return "", err
+		}
+		if statusCode != http.StatusOK {
+			return "", fmt.Errorf("can't clean SANs values for certificate on server side")
+		}
+	}
```
arykalin: Is this code covered by tests?

arykalin: Also it's not very clear for me what we're doing here. I think we need to add comment with short description.

PR #90, opened 2020-03-05
Restore defaults for getcred action

tr1ck3r:
The default values for `--client-id` and `--scope` were lost between versions 4.6.0 and 4.7.1


arykalin:
@tr1ck3r we're setting default values on SDK level, that's why I removed this defaults:
https://github.com/Venafi/vcert/blob/master/pkg/venafi/tpp/tpp.go#L41-L42

Though we can set default client-id to vcert-cli for CLI and vcert-go for SDK.

mr-tron:
I agree with making default value in cli for client_ID but not for scope.

tr1ck3r:
@mr-tron the CLI requires a more limited scope than the SDK based on the fact that the CLI only uses a subset of the functionality so the default scope for the CLI should be different.

@arykalin it now seems best for the default client_id to be "vcert-sdk" for the SDK and "vcert-cli" for the CLI.  We'll use "vcert-sdk" for VCert-Python, VCert-Ruby, and VCert-Java.

PR #107, opened 2020-04-28
Expose the TPP ConfigWrite API

wallrj:
This is needed for a cert-manager extension that we're writing for Venafi.

We need a way to trigger TPP certificiate provisioning by setting the following config:

```
		err = tppClient.ConfigWrite(&tpp.ConfigWriteRequest{
			ObjectDN: response.CertificateDN,
			AttributeData: []tpp.ConfigAttribute{
				{
					Name:  "Stage",
					Value: []string{"750"},
				},
				{
					Name:  "Work To Do",
					Value: []string{"1"},
				},
				{
					Name:  "Management Type",
					Value: []string{"Provisioning"},
				},
			},
		})

```

This API seems to be specific to TPP so I have not added this to the general `endpoint.Connector` interface.

If you agree that this is a useful method to expose, I will add tests.

Alterntatively, I could make the  interface more generic, e.g. `tpp.TriggerProvisioning(certificiateDN string)`

/cc @carlskii @KBo0o

### DIFF
--- a/pkg/venafi/tpp/tpp.go
+++ b/pkg/venafi/tpp/tpp.go
@@ -34,6 +34,7 @@ import (
 
 	"github.com/Venafi/vcert/pkg/certificate"
 	"github.com/Venafi/vcert/pkg/endpoint"
+	"github.com/Venafi/vcert/pkg/verror"
 )
 
 const defaultKeySize = 2048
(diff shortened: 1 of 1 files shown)


KBo0o:
@wgoulet @tr1ck3r How to design for DevOpsACCELERATE compatibility, not just Venafi Platform?

tr1ck3r:
@wallrj we are trying our best to avoid introducing dependencies to VCert that require low-level Config methods like `POST Config/Write` because that would significantly broaden the scope of the API Application access required.  I think that @carlskii and @KBo0o filled me in on this use case last week and we discussed an alternative approach (Bulk Provisioning) that would more in line with our product strategy.  I'll circle back with them.

wallrj:
@tr1ck3r I'll close this and wait to hear more about the Bulk Provisioining API.
Meanwhile we'll maintain a fork at https://github.com/jetstack/vcert/tree/cert-manager-venafi-key-sync

PR #110, opened 2020-05-13
Wrap all Authenticate errors with verror.AuthError

wallrj:
This makes it easier for users of the vcert library to distinguish between
temporary and permanent error conditions.

Signed-off-by: Richard Wall <richard.wall@jetstack.io>


tr1ck3r:
@wallrj Thank you for contributing to the VCert project!

PR #111, opened 2020-05-13
Wrap URL errors from tpp.NewConnector with verror.UserDataError

wallrj:
Allows a vcert user to distinguish between errors caused by bad configuration
and temporary errors which may be resolved by retrying.

I also removed getURL and associated tests.
`baseURL` (a private field) will never be the empty because it is validated earlier in
`NewConnector`  via `normalizeURL`.


tr1ck3r:
@wallrj Thank you for contributing to the VCert project!

PR #116, opened 2020-07-14
Changes to the fake connector to allow testing issue and revoke workflows without a real Venafi instance

BrandonRoehl:
These were changes we made for our CI pipeline to get full coverage and tests around the Venafi integration in workflows where revocation based on the fingerprint is necessary. Willing for a lot of changes and updates to this since we may have varying coding practices.


tr1ck3r:
@BrandonRoehl Thank you for your contribution.  I think some refactoring may be needed.  The changes you've proposed have broken the asynchronous pickup use case for test mode.  An easy way to see this is using the VCert CLI like this:

```
vcert enroll --test-mode --no-prompt --no-pickup --cn my.fake.test --pickup-id-file id.txt
vcert pickup --test-mode --pickup-id-file id.txt
```

This statelessness was accomplished by base64 encoding the serialized **fakeRequestID** that includes the CSR and returning that as the Pickup ID.  With the CSR, the `pickup` action can generate a certificate, albeit a different certificate with each invocation rather than the same certificate as it would be for TPP and Venafi Cloud.  That base64 encoding of the **fakeRequestID** is still happening but the `pickup` result is _"no cert has been issued with this id or thumbprint"_ error rather than a certificate generated from the CSR.

The behavior for `renew` and `revoke` in test mode has also changed (expected), but they've changed in such a way that they're still effectively unsupported but they're now returning errors that don't communicate that: 

- _"renew is not supported in -test-mode"_ ➡️ _"no cert has been issued with this id or thumbprint"_
- _"revocation is not supported in -test-mode"_ ➡️ _"no cert has been issued with this thumbprint"_

It should be possible for the renew case to work since the CSR will provide it with the subject DN, SANs, and key properties needed to generate a new key pair, CSR (for a new pickup ID), and certificate (assuming `--no-pickup` is not specified).  

I think the best way to handle the revocation case is to always return a success result unless the ID or thumbprint indicates it should fail (e.g. the CSR extracted from the ID has a common name that contains the string "no-revoke", or the thumbprint provided is 40 f's "ffffffffffffffffffffffffffffffffffffffff").

BrandonRoehl:
Thanks for the speedy reply. This is something I totally looked over.

Will be looking at fixing this. Are there some tests that we should be conforming for and where should we be running some of this stuff?

BrandonRoehl:
Closing for now will re open when the changes are done

Issue #120, opened 2020-07-30
VCert Revoke with --no-retire (flag needs an argument: -no-retire)


smeluski:
**PROBLEM SUMMARY**
Looking to revoke a certificate without disabling it.

**STEPS TO REPRODUCE**
vcert revoke -tpp-url "https://venafiaddress.com/vedsdk" -tpp-user "***" -tpp-password "***" -id "\\VED\Policy\\Certificates\\Non-Production\\Dev Ops\\certdomain.com" --reason superseded --no-retire

**EXPECTED RESULTS**
cert is revoked, but not disabled

**ACTUAL RESULTS**
error - flag needs an argument: -no-retire

**ENVIRONMENT DETAILS**
vcert version v4.9.6+895

**COMMENTS/WORKAROUNDS**
The help seems to indicate I may need to provide a value for --no-retire, but I am not sure what that would be. I have been able to revoke a cert without --no-retire which causes it to be disabled.

NAME:
   vcert.exe revoke - To revoke a certificate

   --id value               Use to specify the ID of the certificate. Required unless --thumbprint is specified. For revocation,marks the certificate as disabled so that no new certificate can be enrolled to replace it. If a replacement certificate will be enrolled, also specify --no-retire=true.
   --no-retire value        Do not disable certificate object. Works only with --id <certificate DN>
   --reason value           The revocation reason. Options include:
                            "none",
                            "key-compromise",
                            "ca-compromise",
                            "affiliation-changed",
                            "superseded",
                            "cessation-of-operation"

PR #121, opened 2020-07-30
Make --no-retire option work for revoke action

tr1ck3r:
Fix for https://github.com/Venafi/vcert/issues/120


rvelaVenafi:
Code works as expected, the --no-retire flag allows for the revoked certificate to be enrolled again.

Issue #115, opened 2020-07-09
Check environment variables for configuration settings when not specified by other means


tr1ck3r:
**BUSINESS PROBLEM**
Using environment variables is a common way to provide configuration settings to command line utilities and scripts, especially when they are executed as part of a CI/CD pipeline.

**PROPOSED SOLUTION**
When the URL, zone, token, API key, and/or trust bundle are not specified on the command line or by an "ini" file, use values of respective environment variables if assigned (e.g. VCERT_URL, VCERT_ZONE, VCERT_TOKEN, VCERT_APIKEY, VCERT_TRUST_BUNDLE)

**CURRENT ALTERNATIVES**
Specify configuration and credential settings on the command line using `-u`, `-z`, `-t`, `-k`, and/or `--trust-bundle`, or in an "ini" file specified using the `--config` parameter.

PR #124, opened 2020-07-31
Fix for:  #115 Check environment variables for configuration settings when not specified by other means

angelmoo:
This change are for supporting environment variables for most common arguments on vcert, as they are described on the ticket, for fixing this viper dependency were added for retrieving the environment variables.
This changes also include a set of test for testing the new functionality
This changes also include a change on examples/simple-cli/main_test.go

all test set for Vcert passed.


cmd/vcert/config.go
```diff
@@ -41,6 +41,14 @@ func buildConfig(c *cli.Context, flags *commandFlags) (cfg vcert.Config, err err
 		var connectorType endpoint.ConnectorType
 		var baseURL string
 		var auth = &endpoint.Authentication{}
+
+		//case when access token can come from enviroment variable.
+		ttpTokenS := flags.tppToken
```
tr1ck3r, line 46: Typo, variable name should be `tppTokenS`

cmd/vcert/main_test.go
```diff
@@ -763,3 +763,138 @@ func TestWrapArgumentDescription(t *testing.T) {
...
+
+	setEnvironmentVariablesForTpp()
+
+	context := getCliContext()
+	cfg, err := buildConfig(context, &flags)
+
+	//execute the validations.
+	if err != nil {
+		t.Fatalf("Failed to build vcert config: %s", err)
+		unsetEnvironmentVariables()
```
tr1ck3r, line 853: Is this function call necessary?  Seems unreachable in the current state.

cmd/vcert/main_test.go
```diff
@@ -763,3 +763,138 @@ func TestWrapArgumentDescription(t *testing.T) {
...
+
+	//create a context, thiw will be used on the build config function.
+	context := getCliContext()
+
+	cfg, err := buildConfig(context, &flags)
+
+	//execute the validations.
+	if err != nil {
+		t.Fatalf("Failed to build vcert config: %s", err)
+		unsetEnvironmentVariables()
```
tr1ck3r, line 888: Is this function call necessary?  Seems unreachable in the current state.

cmd/vcert/validators.go
```diff
@@ -468,3 +496,44 @@ func validateRevokeFlags1(commandName string) error {
...
+	if getPropertyFromEnvironment(vCertApiKey) != "" {
+		if flags.apiKey != "" {
+			logger.Println(colorYellow, "Warning: Command line parameter -k has overridden environment variable VCERT_APIKEY", colorReset)
+		}
+
+	}
+
+	if getPropertyFromEnvironment(vCertTrustBundle) != "" {
+		if flags.trustBundle != "" {
+			logger.Println(colorYellow, "Warning: Command line parameter -trust-bundle has overridden environment variable VCERT_TRUST_BUNDLE", colorReset)
```
tr1ck3r, line 534: Although both `-trust-bundle` and `--trust-bundle` are valid, we should use the `--trust-bundle` form in messages since that is meant to be the standard for CLI parameters consisting of more than one character.

cmd/vcert/config.go
angelmoo: fixed on last commit

cmd/vcert/main_test.go
angelmoo: fixed on last commit

cmd/vcert/main_test.go
angelmoo: fixed on last commit

cmd/vcert/validators.go
angelmoo: fixed on the last commit.

angelmoo:
@tr1ck3r @asquith-venafi I've updated the code review with the changes you requested.

PR #125, opened 2020-08-01
GitHub 114

rvelaVenafi:
The follwing set of fixes provides the ability to generate the CSR and private key in json format.
Also fixes an issue with the enroll command and the json format when exported to multiple files (certificate, key, chain). The json structure was not generated correctly.

### DIFF
--- a/cmd/vcert/commands.go
+++ b/cmd/vcert/commands.go
@@ -721,7 +721,7 @@ func generateCsrForCommandGenCsr(cf *commandFlags, privateKeyPass []byte) (priva
 
 func writeOutKeyAndCsr(commandName string, cf *commandFlags, key []byte, csr []byte) (err error) {
 	pcc := &certificate.PEMCollection{}
-	pcc.Certificate = string(csr[:])
+	pcc.CSR = string(csr[:])
 	pcc.PrivateKey = string(key[:])
 
 	result := &Result{
@@ -733,7 +733,7 @@ func writeOutKeyAndCsr(commandName string, cf *commandFlags, key []byte, csr []b
 			ChainOption:  certificate.ChainOptionFromString(cf.chainOption),
 			AllFile:      cf.file,
 			KeyFile:      cf.keyFile,
-			CertFile:     cf.csrFile,
+			CSRFile:      cf.csrFile,
 			ChainFile:    "",
 			PickupIdFile: "",
 			KeyPassword:  cf.keyPassword,
--- a/cmd/vcert/commands.go
+++ b/cmd/vcert/commands.go
@@ -16,7 +16,6 @@ import (
 	"log"
 	"net"
 	"net/http"
-	"os"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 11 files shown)


cmd/vcert/flags.go
```diff
@@ -423,6 +423,15 @@ var (
 		Destination: &flags.omitSans,
 	}
 
+	flagCSRFormat = &cli.StringFlag{
+		Name: "format",
+		Usage: "Generates the Certificate Signing Request in the specified format. Options include: pem | json\n" +
+			"\t\tpem: Generates the CSR in classic pem format to be used as a file.\n" +
```
tr1ck3r, line 429: Should be capitalized PEM in the description (i.e. "classic PEM format")

cmd/vcert/flags.go
```diff
@@ -423,6 +423,15 @@ var (
...
+			"\t\tjson: Generates the CSR in json format, suitable for rest api operations.",
```
tr1ck3r, line 430: Sound be capitalized JSON, REST, and API in the description.  Also, for some reason the formatting of these tabs looks off when I build/execute on Linux.  Please investigate that, too.

cmd/vcert/result_writer.go
```diff
@@ -205,33 +206,57 @@ func (r *Result) Flush() error {
...
-			} else {
-				err = ioutil.WriteFile(r.Config.CertFile, bytes, 0600)
-				errors = append(errors, err)
-			}
+			err = writeFile(certFileOutput, r.Config, r.Config.CertFile)
+			//bytes, err := certFileOutput.Format(r.Config)
+			//if err != nil {
+			//	return err // something worse than file permission problem
+			//} else {
+			//	err = ioutil.WriteFile(r.Config.CertFile, bytes, 0600)
```
tr1ck3r, line 214: Let's remove old code rather than leaving it commented out

cmd/vcert/commands.go
```diff
@@ -720,37 +719,27 @@ func generateCsrForCommandGenCsr(cf *commandFlags, privateKeyPass []byte) (priva
 	return
 }
 
-func writeOutKeyAndCsr(cf *commandFlags, key []byte, csr []byte) (err error) {
+func writeOutKeyAndCsr(commandName string, cf *commandFlags, key []byte, csr []byte) (err error) {
+	pcc := &certificate.PEMCollection{}
+	pcc.Certificate = string(csr[:])
```
tr1ck3r, line 724: We should add "CSR" (omitempty) to the PEMCollection struct so that the data is accurately named when we output this in JSON.  Then this will become `pcc.CSR = string(csr[:])`.

cmd/vcert/flags.go
rvelaVenafi: modified PEM text to be capitalized

cmd/vcert/result_writer.go
rvelaVenafi: Removed unneeded code

cmd/vcert/commands.go
rvelaVenafi: Added CSR attribute to PEMCollection{} object. And updated code to make use of it.

cmd/vcert/flags.go
rvelaVenafi: Capitalized JSON, REST and API keywords

PR #127, opened 2020-08-04
CSR not being output when using --format pem

rvelaVenafi:
The following fix solves an issue where the csr is not being output in console or file when using --format pem

### DIFF
--- a/cmd/vcert/result_writer.go
+++ b/cmd/vcert/result_writer.go
@@ -128,25 +128,17 @@ func (o *Output) Format(c *Config) ([]byte, error) {
 
 	default: // pem
 		res := ""
-		certValue := ""
-		if o.Certificate == "" {
-			if o.CSR != "" {
-				certValue = o.CSR
-			}
-		} else {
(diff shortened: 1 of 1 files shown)


cmd/vcert/result_writer.go
```diff
@@ -128,16 +128,25 @@ func (o *Output) Format(c *Config) ([]byte, error) {
...
-			res += o.Certificate
+			res += certValue
 			res += o.PrivateKey
 		case certificate.ChainOptionIgnore:
-			res += o.Certificate
+			res += certValue
 			res += o.PrivateKey
 		default:
-			res += o.Certificate
+			res += certValue
```
tr1ck3r, line 149: @rvelaVenafi since the ChainOption will never apply to situations involving a CSR, can we address this defect a bit more simply by just doing this here (since string variables in golang are "" by default)?
```
res += o.Certificate
res += o.CSR
```

rvelaVenafi: Added suggested changes.

Issue #134, opened 2020-09-24
VCert CLI argument for CSR is silently ignored when value is invalid


tr1ck3r:
**PROBLEM SUMMARY**
VCert CLI argument for CSR is silently ignored when value is invalid

**STEPS TO REPRODUCE**
```
vcert enroll -u https://tpp.venafi.example -t hw4+yAeOUcjTLao167ptIA== -z VCert --cn test.venafi.example --csr test.req
```

**EXPECTED RESULTS**
An error message stating that the value of the `--csr` parameter must be "local", "service", or "file:<path-to-csr-file>".  If the correct syntax had been specified VCert would have correctly returned the following error message:

> The '--cn' parameter cannot be used with '--csr file:'

**ACTUAL RESULTS**
```
Enter key pass phrase:************
Verifying - Enter key pass phrase:************
vCert: 2019/12/11 07:49:44 Successfully connected to TPP
vCert: 2019/12/11 07:49:48 Successfully read zone configuration for VCert
vCert: 2019/12/11 07:49:49 Successfully created request for test.venafi.example
vCert: 2019/12/11 07:49:50 Successfully posted request for test.venafi.example, will pick up by \VED\Policy\VCert\test.venafi.example
vCert: 2019/12/11 07:49:50 Issuance of certificate is pending...
```
The `--csr` parameter was silently ignored and a certificate was enrolled using local CSR generation (the default behavior when `--csr` is not specified).

**COMMENTS/WORKAROUNDS**
While investigating this I also found we’re not returning an error when `--csr file:` is specified with `--key-type`, `--key-size`, and/or `--key-curve`, they’re just being silently ignored.  When we fix this bug we should add a check for that condition that return an error similar to the one we return when `--csr file:` is specified with `--cn`.

wgoulet:
@tr1ck3r FYI on a whim I tried to reproduce this issue with a fresh build of vcert from master and when I executed these exact repro steps I did not see the enrollment occur:

`wgoulet@DESKTOP-O26JR0I:~/go/src/github.com/wgoulet/vcert$ ./bin/linux/vcert enroll -u https://example.venafi.com -t XXXXXXXXXXX -z VCert\\wgoulet --cn test.venafi.example --csr test.csr --trust-bundle venqa.pem
vCert: 2020/09/25 00:55:51 unexpected -csr option: test.csr
wgoulet@DESKTOP-O26JR0I:~/go/src/github.com/wgoulet/vcert$`

[...]

PR #136, opened 2020-09-26
Fixes for issue #134 and fixing tests

wgoulet:
This PR attempts to address the issues reported in #134 . For this fix, I also included some minor updates to the docs to indicate all the environment variables that should be set when running 'make tpp_test'. I also added some new tests for the cmd package and added an option to the makefile to be able to run those tests with make. Finally, I fixed a broken test in the cmd package where the path separator was missing from a call to create a local file.

I am not setup to run cross-platform tests across all 3 platforms and the only test I am unable to execute successfully is Test_GetCertificateList in pkg/venafi/tpp due to not having a system with a large enough list of certificates in a single policy folder to test against. Other than these comments, all TPP tests pass on Linux as well as all cmd/vcert tests pass.

### DIFF
--- a/cmd/vcert/validators.go
+++ b/cmd/vcert/validators.go
@@ -59,9 +59,15 @@ func validateCommonFlags(commandName string) error {
 		return fmt.Errorf("The '-file' option cannot be used used with any other -*-file flags. Either all data goes into one file or individual files must be specified using the appropriate flags")
 	}
 
-	csrOptionRegex := regexp.MustCompile(`^file:.*$|^local$|^service$|^$`)
+	csrOptionRegex := regexp.MustCompile(`(^file:).*$|^local$|^service$|^$`)
 	if !csrOptionRegex.MatchString(flags.csrOption) {
-		return fmt.Errorf("unexpected -csr option provided: %s; specify one of the following options: %s, %s, or %s.", flags.csrOption, "'file:<filename>'", "'local'", "'service'")
+		return fmt.Errorf("unexpected -csr option provided: %s; specify one of the following options: %s, %s, or %s", flags.csrOption, "'file:<filename>'", "'local'", "'service'")
+	}
+
+	csrOptFlagResults := csrOptionRegex.FindStringSubmatch(flags.csrOption)
+
+	if csrOptFlagResults[1] != "" && (flags.keyTypeString != "" || flags.keyCurveString != "" || flags.keySize > 0) {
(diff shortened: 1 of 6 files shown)


cmd/vcert/validators.go
```diff
@@ -58,9 +59,15 @@ func validateCommonFlags(commandName string) error {
 		return fmt.Errorf("The '-file' option cannot be used used with any other -*-file flags. Either all data goes into one file or individual files must be specified using the appropriate flags")
 	}
 
-	csrOptionRegex := regexp.MustCompile(`^file:.*$|^local$|^service$|^$`)
+	csrOptionRegex := regexp.MustCompile(`(^file:).*$|^local$|^service$|^$`)
 	if !csrOptionRegex.MatchString(flags.csrOption) {
-		return fmt.Errorf("unexpected -csr option: %s", flags.csrOption)
+		return fmt.Errorf("unexpected -csr option provided: %s; specify one of the following options: %s, %s, or %s", flags.csrOption, "'file:<filename>'", "'local'", "'service'")
```
tr1ck3r, line 64: Even though we still allow the single dash for backward compatibility, the double dash is the proper syntax so should be `--csr` here.

cmd/vcert/validators.go
```diff
@@ -58,9 +59,15 @@ func validateCommonFlags(commandName string) error {
...
+	}
+
+	csrOptFlagResults := csrOptionRegex.FindStringSubmatch(flags.csrOption)
+
+	if csrOptFlagResults[1] != "" && (flags.keyTypeString != "" || flags.keyCurveString != "" || flags.keySize > 0) {
+		return fmt.Errorf("the '-keytype','-keycurve' and 'key-size' options cannot be used when '-csr file:' option is provided")
```
tr1ck3r, line 70: Should be `--key-type`, `--key-curve`, `--key-size`, and `--csr`.

wgoulet: @tr1ck3r Updated the error messages with the -- as requested.

tr1ck3r (approved):
Changes look good to me

rvelaVenafi (approved):
Very nice test you did here. Thanks a lot.

angelmoo (approved):
code looks good

rvelaVenafi (approved):
Changes look good. 
Thanks for your contribution!

wgoulet:
@rvelaVenafi Thanks I screwed up in github UI and accidentally set you to re-review; pls ignore that.

rvelaVenafi:
> @rvelaVenafi Thanks I screwed up in github UI and accidentally set you to re-review; pls ignore that.

No worries!

PR #149, opened 2021-01-18
Vcert jks support

marcos-albornoz:
This PR is for the code to provide the support to format the output of the certificate, private key and the chain of trust to JKS similar to the behavior when the format to export is PKCS12.
In accordance of the Java Key Store specification, it's required a password to protect the Java Key Store, also it's required a password to protect the entry that will contain the certificate, private key and the chain of trust, and an alias for that entry.

So given the previous explanation, following the rules which are implemented in the current PR:

- User requests a **Java Key Store** by specifying _"jks"_ after the _-format_ switch

- User must use the _-file_ switch to specify the name of the keystore file when they specify _-format jks_.

- User specifies the password for the **store file and the key entry**( conformed by the private key, the certificate and the chain certificates) using the _-jks-password_(optionally) switch and the **key password**(_-key-password_ switch or pass phrase from prompt) respectively. If the _-jks-password_ is not provided, then the value of _key password_ is used for that.

- User always will be requested for passwords of at least 6 characters so the _-no-prompt_ switch only can be used if the _-key-password_ switch is specified.

- User must use the _-jks-alias_ switch to provide the alias for the key entry.

In summary, the rules existing for PKCS12 format applies for JKS format, with the difference that the passwords and the alias are mandatory.

Following a valid enroll declaration: 
`vcert enroll -u https://supertreat.venqa.venafi.com/ -t jhKi1xFxM5UmClnXsvaAoA== -cn amoo69.test.venafi.example.com -z devops -format jks -file test.jks -key-password abcdef -jks-alias alias1 -jks-password 123456`

### DIFF
--- a/aruba/features/format/jks.feature
+++ b/aruba/features/format/jks.feature
@@ -52,6 +52,10 @@ Feature: JKS format output
       When I enroll random certificate in test-mode with -no-prompt -format jks -file all.jks -jks-alias alias
         Then it should fail with "JKS format requires passwords that are at least 6 characters long"
 
+  Scenario: where JKS format is specified but a key-password is not provided
+      When I enroll random certificate in test-mode with -no-prompt -format jks -file all.jks -jks-password 123456 -jks-alias alias
+        Then it should fail with "JKS format requires passwords that are at least 6 characters long"
+
   Scenario: where JKS format is not specified but the jks password is provided
     When I enroll random certificate in test-mode with -no-prompt -format pkcs12 -jks-password 123456 -file all.jks
       Then it should fail with "The --jks-password parameter may only be used with --format jks"
@@ -61,27 +65,17 @@ Feature: JKS format output
       Then it should fail with "The --jks-alias parameter may only be used with --format jks"
 
   Scenario: where all objects are written to one JKS archive
-    When I enroll random certificate in test-mode with -no-prompt -format jks -file all.jks -jks-password 123456 -jks-alias xxx
+    When I enroll random certificate in test-mode with -no-prompt -format jks -file all.jks -key-password 123456 -jks-password 123456 -jks-alias abc
     Then the exit status should be 0
     And "all.jks" should be jks archive with password "123456"
 
   Scenario: where all objects are written to one JKS archive
-      When I enroll random certificate in test-mode with -no-prompt -format jks -file all.jks -jks-password 123456 -jks-alias xxx -key-type ecdsa
+      When I enroll random certificate in test-mode with -no-prompt -format jks -file all.jks -key-password 123456 -jks-password 123456 -jks-alias abc -key-type ecdsa
       Then the exit status should be 0
       And "all.jks" should be jks archive with password "123456"
 
-  Scenario Outline: where all objects are written to one JKS archive providing with key-password
-    When I enroll random certificate in <endpoint> with -format jks -file all.jks -key-password 123456 -jks-alias xxx
-    Then the exit status should be 0
-    And "all.jks" should be jks archive with password "123456"
-    Examples:
-       | endpoint  |
-       | test-mode |
-       | TPP       |
-       | Cloud     |
-
   Scenario Outline: where all objects are written to one JKS archive with key-password and providing the jks-password
-    When I enroll random certificate in <endpoint> with -format jks -file all.jks -key-password 123abc -jks-password 123456 -jks-alias xxx
+    When I enroll random certificate in <endpoint> with -format jks -file all.jks -key-password 123abc -jks-password 123456 -jks-alias abc
     Then the exit status should be 0
     And "all.jks" should be jks archive with password "123456"
     Examples:
@@ -130,7 +124,7 @@ Feature: JKS format output
 
   Scenario Outline: where it pickups up service-generated certificate and outputs it in JKS format
     When I enroll random certificate using <endpoint> with -no-prompt -no-pickup -csr service
-    And I retrieve the certificate using <endpoint> using the same Pickup ID with -timeout 180 -key-password newPassw0rd! -file all.jks -format jks -jks-alias xxx
+    And I retrieve the certificate using <endpoint> using the same Pickup ID with -timeout 180 -key-password newPassw0rd! -file all.jks -format jks -jks-alias abc
     And "all.jks" should be JKS archive with password "newPassw0rd!"
     Examples:
       | endpoint  |
@@ -140,7 +134,7 @@ Feature: JKS format output
 
 #  Scenario Outline: Pickup JKS with typing pass phrases
 #    When I enroll random certificate using <endpoint> with -no-prompt -no-pickup -csr service
-#    And I interactively retrieve the certificate using <endpoint> using the same Pickup ID with -timeout 99 -file all.jks -format jks -jks-alias xxx
+#    And I interactively retrieve the certificate using <endpoint> using the same Pickup ID with -timeout 99 -file all.jks -format jks -jks-alias abc
 #    And I type "newPassw0rd!"
 #    And I type "newPassw0rd!"
 #    Then the exit status should be 0
@@ -148,8 +142,8 @@ Feature: JKS format output
 #    Examples:
 #      | endpoint  |
 #      | test-mode |
-      # | TPP       |
-      # | Cloud     | # -csr service is not supported by Cloud
+#      | TPP       |
+#      | Cloud     | # -csr service is not supported by Cloud
 
 #  Scenario Outline: where it should enroll a JKS certificate with -csr service and without file option (VEN-48622)
 #    When I enroll random certificate using <endpoint> with -csr service -no-prompt -no-pickup -format pkcs12
--- a/aruba/features/format/jks.feature
+++ b/aruba/features/format/jks.feature
@@ -37,14 +37,20 @@ Feature: JKS format output
       Then it should fail with "JKS format requires certificate, private key, and chain to be written to a single file; specify using --file"
 
   Scenario: where JKS format is specified, but a short password is used
-    When I enroll random certificate in test-mode with -no-prompt -format jks -key-password 1234 -file all.jks
+    When I enroll random certificate in test-mode with -no-prompt -format jks -key-password 1234 -jks-password 123456 -file all.jks
       Then it should fail with "JKS format requires passwords that are at least 6 characters long"
-    When I enroll random certificate in test-mode with -no-prompt -format jks -jks-password 1234 -file all.jks
+    When I enroll random certificate in test-mode with -no-prompt -format jks -key-password 123456 -jks-password 1234 -file all.jks
       Then it should fail with "JKS format requires passwords that are at least 6 characters long"
+    When I enroll random certificate in test-mode with -no-prompt -format jks -key-password 1234 -jks-password 1234 -file all.jks
+          Then it should fail with "JKS format requires passwords that are at least 6 characters long"
 
   Scenario: where JKS format is specified and a password is used but the jks alias is not provided
     When I enroll random certificate in test-mode with -no-prompt -format jks -key-password 123456 -file all.jks
       Then it should fail with "The --jks-alias parameter is required with --format jks"
+
+  Scenario: where JKS format is specified but a password is not provided
+      When I enroll random certificate in test-mode with -no-prompt -format jks -file all.jks -jks-alias alias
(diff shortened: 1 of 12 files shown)


cmd/vcert/result_writer.go
```diff
@@ -195,6 +276,11 @@ func (r *Result) Flush() error {
 			if err != nil {
 				return fmt.Errorf("failed to encode pkcs12: %s", err)
 			}
+		} else if r.Config.Format == "jks" {
```
angelmoo, line 279: I think as this value is being used in more than 2 places, we should convert it on a constant variable.

tr1ck3r (commented):
Great work @marcos-albornoz.  I've committed some updates to error messages and help text to your branch as we discussed.  The password handling isn't quite the behavior I was expecting which is:

1. When `--format jks`, a key password is required and is specified by either `--key-password` or the password prompt when `--key-password` is omitted.
2. The key password value is always used for the **PrivateKeyEntry** password.
3. If `--jks-password` is specified, then its value is used for the **Store** password.
4. If `--jks-password` is not specified, then the key password value is used for the **Store** password.
5. The passwords must be at least 6 characters long (third-party requirement).

cmd/vcert/result_writer.go
marcos-albornoz: done!

cmd/vcert/validators.go
```diff
@@ -176,6 +170,58 @@ func validatePKCS12Flags(commandName string) error {
...
+		}
+		if (flags.csrOption == "" || flags.csrOption == "local") && flags.noPickup {
+			return fmt.Errorf(`JKS format is not allowed for the enroll or renew actions when -csr is "local" and -no-pickup is specified`)
+		}
+
+		if flags.jksPassword != "" {
+			if len(flags.jksPassword) < JKSMinPasswordLen {
+				return fmt.Errorf("Password for JKS format must be at least %d characters (see --jks-password)", JKSMinPasswordLen)
+			}
+		} else if flags.keyPassword != "" {
```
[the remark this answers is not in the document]
marcos-albornoz: done!

cmd/vcert/result_writer.go
```diff
@@ -117,6 +122,82 @@ func (o *Output) AsPKCS12(c *Config) ([]byte, error) {
...
+
+	//setting as jksPassword the value in --jks-password (if it was provided) or the value from keyPassword( --key-password or pass phrase value)
+	if c.JKSPassword != "" {
+		jksPassword = []byte(c.JKSPassword)
+	} else {
+		jksPassword = []byte(c.KeyPassword)
+	}
+
+	//adding the PK entry to the JKS
+	if err := keyStore.SetPrivateKeyEntry(c.JKSAlias, pkeIn, jksPassword); err != nil {
```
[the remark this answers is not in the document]
marcos-albornoz: done!

cmd/vcert/result_writer.go
```diff
@@ -173,24 +173,33 @@ func (o *Output) AsJKS(c *Config) ([]byte, error) {
...
 
-	var jksPassword []byte
+	var keyPass []byte
 
-	//setting as jksPassword the value in --jks-password (if it was provided) or the value from keyPassword( --key-password or pass phrase value)
+	//setting as keyPass the value from keyPassword(--key-password or pass phrase value) or the value in --jks-password
+	if c.KeyPassword != "" {
+		keyPass = []byte(c.KeyPassword)
+	} else {
+		keyPass = []byte(c.JKSPassword)
```
tr1ck3r, line 182: We always want `keyPass = []byte(c.KeyPassword)`, never `keyPass = []byte(c.JKSPassword)` for consistency with other formats (i.e. in VCert the `--key-password` is always focused on protecting the private key).

marcos-albornoz: mmmm… but according the chat that we had in the related email, we agreed that when it isn’t provided the private key password, that is when the --no-prompt flag is provided and the -key-password is not provided, then we can use the -jks-password value for both, the store password and also for the key entry password, and that code is for that case… or am I missing something?

cmd/vcert/result_writer.go
```diff
@@ -117,6 +122,91 @@ func (o *Output) AsPKCS12(c *Config) ([]byte, error) {
...
+
+	//creating a PK entry
+	pkeIn := keystore.PrivateKeyEntry{
+		CreationTime: time.Now(),
+		PrivateKey:   privDER,
+		CertificateChain: []keystore.Certificate{
+			{
+				Type:    "X509",
+				Content: chainCertInBytes,
+			},
```
tr1ck3r, line 172: I think this may be expecting one `keystore.Certificate` object per certificate in the chain (EE + all the CAs) rather than just one `keystore.Certificate` with the bytes of all of the certificates in the chain for its `Content`.

cmd/vcert/result_writer.go
tr1ck3r: Yeah, I don't want the value of the `--jks-password` ever being used for the _PrivateKeyEntry_, only for the _Store_.   The `--jks-password` is the optional and when not specified the value of the `--key-password` should be used for the _Store_ which is how that is currently coded.

marcos-albornoz: I understand the point that you mention... the private key entry should be protected only by the key password( or the pass phrase if that was provided by prompt action) and the -jks-password should be used only to provide the password for the key store, but as we commented, what about of the case when neither the -key-password and the pass phrase were not provided( --no-prompt)? The password for the key entry is mandatory, at least it is a limit given the keystore go library which we are using

tr1ck3r: Yes, that's right.  If a user specifies `--no-prompt` without also specifying `--key-password` with `--format jks` the expected result is an error stating passwords must be at least 6 characters long (because the default key password value is the empty string).

marcos-albornoz: got it, in that way the user is required to provide a password for the key entry... I will do the change :-)

tr1ck3r: Thank you @marcos-albornoz 😃

cmd/vcert/result_writer.go
marcos-albornoz: you're correct, I've update that code to add each certificate individually instead of the chain of their bytes

cmd/vcert/result_writer.go
marcos-albornoz: done!

cmd/vcert/result_writer.go
marcos-albornoz: done!

tr1ck3r (approved):
Great job @marcos-albornoz 😃 Code looks good to me now and the verifications I've done with a binary built from your branch have all been successful.

angelmoo (approved):
Changes looks good

aruba/features/format/jks.feature
```diff
@@ -0,0 +1,169 @@
...
+
+  - User requests JKS by specifying "jks" after the -format switch
+
+  - User must use the -file switch to specify the name of the keystore file when they specify -format jks
+    (i.e. neither the -cert-file, -key-file, nor -chain-file switches may appear on the command line,
+    and console output as a base64 encoded blob will not be supported).
+
+  - User always will be request for a password of at least 6 characters when he request a JKS file so it isn't possible to use the -no-prompt switch
+
+  - User specifies the password for the JKS file and the key entry, conformed by the private kye, the certificate and the chain certificates, using, preferable,
```
rvelaVenafi, line 15: Minor typo at "private kye". It must be "private key"

aruba/features/format/jks.feature
```diff
@@ -0,0 +1,169 @@
...
+    the -jks-password or alternatively the -key-password switch
+
+  - User must use the -jks-alias switch to provide the alias for the jks entry that will conformed by the private kye, the certificate and the chain certificates
```
rvelaVenafi, line 18: Minor typo at "jks entry that will conformed by"...it must be "jks entry that will be conformed by "

rvelaVenafi: Also typo o "kye" word. Must be "key"

aruba/features/format/jks.feature
```diff
@@ -0,0 +1,169 @@
...
+    When I enroll random certificate using <endpoint> with -no-prompt -no-pickup -csr service
+    And I retrieve the certificate using <endpoint> using the same Pickup ID with -timeout 180 -key-password newPassw0rd! -file all.jks -format jks -jks-alias abc
+    And "all.jks" should be JKS archive with password "newPassw0rd!"
+    Examples:
+      | endpoint  |
+      | test-mode |
+      | TPP       |
+      # | Cloud     | # -csr service is not supported by Cloud
+
+#  Scenario Outline: Pickup JKS with typing pass phrases
```
rvelaVenafi, line 146: If these code is no longer needed, why not remove it?

cmd/vcert/result_writer.go
```diff
@@ -117,6 +122,106 @@ func (o *Output) AsPKCS12(c *Config) ([]byte, error) {
 	return bytes, nil
 }
 
+func (o *Output) AsJKS(c *Config) ([]byte, error) {
+
+	var err interface{}
```
rvelaVenafi, line 127: you dont need to declare err here. You can do it inline as 
var, err := myCode()

cmd/vcert/result_writer.go
```diff
@@ -117,6 +122,106 @@ func (o *Output) AsPKCS12(c *Config) ([]byte, error) {
...
+			Content: crt.Bytes,
+		})
+	}
+
+	// getting the bytes of the PK
+	p, _ = pem.Decode([]byte(o.PrivateKey))
+	if p == nil {
+		return nil, fmt.Errorf("missing private key PEM")
+	}
+	var privDER []byte
```
rvelaVenafi, line 162: the same with privDER, declaration not needed

cmd/vcert/result_writer.go
```diff
@@ -117,6 +122,106 @@ func (o *Output) AsPKCS12(c *Config) ([]byte, error) {
...
+	//decrypting the PK because due the restriction that always will be requested the key password
+	//to the user(--key-password or pass phrase value from prompt) for jks format then the PK always
+	//will be encrypted with the key password provided
+	privDER, err = x509.DecryptPEMBlock(p, []byte(c.KeyPassword))
+	if err != nil {
+		return nil, fmt.Errorf("private key PEM decryption error: %s", err)
+	}
+
+	//Unmarshalling the PK
+	var privKey interface{}
```
rvelaVenafi, line 173: The same with privKey, declaration not needed

cmd/vcert/utils.go
```diff
@@ -44,6 +44,8 @@ const (
 	/* #nosec */
 	vCertApiKey      = "VCERT_APIKEY"
 	vCertTrustBundle = "VCERT_TRUST_BUNDLE"
+
+	JKSFormat = "jks"
```
rvelaVenafi, line 48: shouldn't this be a constant?

rvelaVenafi (changes requested):

aruba/features/format/jks.feature
marcos-albornoz: fixing

aruba/features/format/jks.feature
marcos-albornoz: fixing

aruba/features/format/jks.feature
marcos-albornoz: fixing

cmd/vcert/result_writer.go
marcos-albornoz: yes, that declaration is a legancy for the previous changes

cmd/vcert/result_writer.go
marcos-albornoz: the same answer as the previous one

cmd/vcert/result_writer.go
marcos-albornoz: the same answer as the previous one :-P

cmd/vcert/utils.go
marcos-albornoz: ahhh... but it is in a const block

rvelaVenafi: My bad, it is indeed in a constant block. All good

rvelaVenafi (approved):

Issue #151, opened 2021-01-22
Ability to revoke OAUTH token using vcert cli


faisalrazzak:
**BUSINESS PROBLEM**

Due to different compliance or security requirements customers may need to revoke/delete OAUTH tokens once different tasks associated with TPP have completed. OAUTH tokens are generally issued for longer during and the idea is that re-authentication of credentials is minimized, however, customers may have shorter during tasks as well where revoking a token after the task is finished will make more sense.

**PROPOSED SOLUTION**

Introduce a new sub-command **token** and a parameter **revoke/delete**. This will remove the issued OAUTH token. The output of this command should be **true** or **false**, something that can be easily consumed in pipelines and chained with other tasks in pipelines.

`vcert token --delete`

**or**

`vcert token --revoke`

**CURRENT ALTERNATIVES**

N/A

tr1ck3r:
https://github.com/Venafi/vcert/tree/tpp-cred-mgmt

Issue #150, opened 2021-01-22
Ability to check validity of OAUTH token using vcert cli


faisalrazzak:
**BUSINESS PROBLEM**

Teams managing security credentials for different vendor applications through CI/CD pipelines also need the ability to constantly check the validity of tokens. This can help in building automated workflows where if a OAUTH token has expired, a new OAUTH token is automatically requested via Refresh Tokens **(or)** stored credentials.

**PROPOSED SOLUTION**

Introduce a new sub-command **token** and a parameter **checkvalidity**. The output of this command should be **true** or **false**, something that can be easily consumed in pipelines and chained with other tasks in pipelines.

`vcert token --checkvalidity` 

**CURRENT ALTERNATIVES**

N/A

tr1ck3r:
https://github.com/Venafi/vcert/tree/tpp-cred-mgmt

PR #153, opened 2021-01-24
Adding checkcred and voidcred actions to VCert CLI

tr1ck3r:
This PR adds the functionality requested by https://github.com/Venafi/vcert/issues/150 and https://github.com/Venafi/vcert/issues/151 via new `checkcred` and `voidcred` actions.

```
# ./vcert getcred -u https://tpp.venafi.example --username local:admin --password somePassw0rd!
vCert: 2021/01/24 23:01:21 Getting credentials...
access_token:  KIKro8hL/4EvvSeVuJQVlw==
access_token_expires:  2021-04-24T23:01:23Z
refresh_token:  PqIQ+4KAEFHGkHZM95b3og==
#
# ./vcert checkcred -u https://tpp.venafi.example -t KIKro8hL/4EvvSeVuJQVlw==
vCert: 2021/01/24 23:01:43 Checking credentials...
access_token_expires:  2021-04-24T23:01:22Z
grant_expires:  2022-01-24T23:01:22Z
client_id:  vcert-cli
scope:  certificate:manage,revoke
#
# ./vcert voidcred -u https://tpp.venafi.example -t KIKro8hL/4EvvSeVuJQVlw==
vCert: 2021/01/24 23:02:02 Access token grant successfully revoked
```

### DIFF
--- a/cmd/vcert/commands.go
+++ b/cmd/vcert/commands.go
@@ -415,7 +415,9 @@ func doCommandCredMgmt1(c *cli.Context) error {
 				tm := time.Unix(int64(resp.Expires), 0).UTC().Format(time.RFC3339)
 				fmt.Println("access_token: ", resp.Access_token)
 				fmt.Println("access_token_expires: ", tm)
-				fmt.Println("refresh_token: ", resp.Refresh_token)
+				if resp.Refresh_token != "" {
+					fmt.Println("refresh_token: ", resp.Refresh_token)
+				}
 			}
 		} else {
 			return fmt.Errorf("Failed to determine credentials set")
@@ -443,9 +445,7 @@ func doCommandCredMgmt1(c *cli.Context) error {
 				iso8601fmt := "2006-01-02T15:04:05Z"
 				tm, _ := time.Parse(iso8601fmt, resp.AccessIssuedOn)
 				accessExpires := tm.Add(time.Duration(resp.ValidFor) * time.Second).Format(iso8601fmt)
-				fmt.Println("access_token_issued: ", resp.AccessIssuedOn)
 				fmt.Println("access_token_expires: ", accessExpires)
-				fmt.Println("grant_issued: ", resp.GrantIssuedOn)
 				fmt.Println("grant_expires: ", resp.Expires)
 				fmt.Println("client_id: ", resp.ClientID)
 				fmt.Println("scope: ", resp.Scope)
--- a/cmd/vcert/commands.go
+++ b/cmd/vcert/commands.go
@@ -526,7 +526,7 @@ func doCommandVoidCred1(c *cli.Context) error {
 		if err != nil {
 			return err
 		}
-		logf("Access token successfully revoked")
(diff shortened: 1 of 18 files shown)


angelmoo (approved):
Changes looks good

cmd/vcert/commands.go
```diff
@@ -318,73 +346,127 @@ func doCommandGetcred1(c *cli.Context) error {
...
+				jsonData, err := json.MarshalIndent(resp, "", "    ")
+				if err != nil {
+					return err
+				}
+				fmt.Println(string(jsonData))
+			} else {
+				tm := time.Unix(int64(resp.Expires), 0).UTC().Format(time.RFC3339)
+				fmt.Println("access_token: ", resp.Access_token)
+				fmt.Println("access_token_expires: ", tm)
+				fmt.Println("refresh_token: ", resp.Refresh_token)
```
marcos-albornoz, line 375: is it not needed to do a nil validation like is in line 396 and 418?

cmd/vcert/commands.go
```diff
@@ -318,73 +346,127 @@ func doCommandGetcred1(c *cli.Context) error {
...
+				}
+				fmt.Println(string(jsonData))
+			} else {
+				tm := time.Unix(int64(resp.Expires), 0).UTC().Format(time.RFC3339)
+				fmt.Println("access_token: ", resp.Access_token)
+				fmt.Println("access_token_expires: ", tm)
+				if resp.Refresh_token != "" {
+					fmt.Println("refresh_token: ", resp.Refresh_token)
+				}
+			}
```
marcos-albornoz, line 399: can this block of code be converted to a common func? I see that is used in more than one place.

cmd/vcert/commands.go
rvelaVenafi: isnt line 356 the nil validation?

cmd/vcert/commands.go
tr1ck3r: The `else` block varies by which action is specified (`getcred`, `checkcred`, and `voidcred`).  The code for the JSON output is common but it is only 5 lines on code (3 for checking error) so I didn't think it was necessary.  Do you think so?

cmd/vcert/commands.go
marcos-albornoz: that is right :-)

rvelaVenafi (approved):

cmd/vcert/commands.go
marcos-albornoz: yes, I though the same, but also I though that if the code for the output changes/increment, then could be more easy to do the change only in one place reducing the risk to forgot to do the change in the places where the code logic is common. that is why I'm asking about if you consider to do that change :-)

marcos-albornoz (approved):

cmd/vcert/commands.go
tr1ck3r: Added `outputJSON` function to address

PR #163, opened 2021-03-31
added the new parameters TPP_IP and TPP_CN 

marcos-albornoz:
added the new parameters TPP_IP and TPP_CN to the docker run command to pass the values of the IP and CN of the virtualized TPP instance.


ricrodriguezg (approved):
Look good to me

luispresuelVenafi (approved):

PR #169, opened 2021-05-07
Rebranding Venafi Cloud to VaaS

tr1ck3r:
Feedback from the release candidate was that the help wasn't clear about whether or not to escape the backslash in the `--z` value.  This is because that behavior varies by operating system and shell.  Given that policy folder DNs in TPP and application names and CIT aliases in VaaS often include spaces, I updated the help to show the zone value being surrounded in double quotes.  I also was more explicit about the actual zone value in the examples since we have them for both TPP and VaaS.


angelmoo (approved):
Changes looks good

rvelaVenafi (approved):

marcos-albornoz (approved):

PR #177, opened 2021-07-16
Ssh cert feature

angelmoo:
added commands:
**sshenroll** for requesting a certificate.
**sshpickup** for  retrieving a certificate.
added test cases for testing the new functionalities.


cmd/vcert/commands.go
```diff
@@ -145,6 +147,24 @@ var (
...
 		vcert getpolicy -k <VaaS API key> -z "<app name>\<CIT alias>"`,
 	}
+
+	commandSshPickup = &cli.Command{
+		Before:    runBeforeCommand,
+		Name:      commandSshPickupName,
+		Flags:     sshPickupFlags,
+		Action:    doCommandSshPickup,
+		Usage:     "To retrieve a SSH Certificate",
+		UsageText: `vcert sshpickup -u https://tpp.example.com -t <TPP access token> --pickup-id <ssh cert DN>"`,
```
tr1ck3r, line 157: Trailing double-quote should be removed.

cmd/vcert/commands.go
```diff
@@ -145,6 +147,24 @@ var (
...
+	}
+
+	commandSshEnroll = &cli.Command{
+		Before:    runBeforeCommand,
+		Name:      commandSshEnrollName,
+		Flags:     sshEnrollFlags,
+		Action:    doCommandEnrollSshCert,
+		Usage:     "To enroll a SSH Certificate",
+		UsageText: `vcert sshenroll -u https://tpp.example.com -t <TPP access token> --template "<val>" --id <val> --principal "bob" --principal "alice" --valid-hours 1"`,
```
tr1ck3r, line 166: Trailing double-quote should be removed.

cmd/vcert/commands.go
```diff
@@ -397,11 +592,22 @@ func doCommandCredMgmt1(c *cli.Context) error {
...
 				User:     cfg.Credentials.User,
 				Password: cfg.Credentials.Password,
 				Scope:    flags.scope,
-				ClientId: flags.clientId})
+				ClientId: flags.clientId}
+
+			if flags.sshCred {
+				auth.Scope = "ssh:manage"
+			} else if flags.pmCred {
+				auth.Scope = "configuration:manage"
```
tr1ck3r, line 605: @angelmoo I think this was my fault for telling you wrong... I think policy management requires `"certificate:manage"` for `getpolicy` and `"configuration:manage"` for `setpolicy`, right?  If so, we should get rid of the `--cpm` flag and update the scope for the `--pm` flag to `"certificate:manage,revoke;configuration:manage"`.

cmd/vcert/validators.go
```diff
@@ -691,3 +693,86 @@ func validateSetPolicyFlags(commandName string) error {
...
+
+func validateSshEnrollFlags(commandName string) error {
+
+	err := validateConnectionFlags(commandName)
+	if err != nil {
+		return err
+	}
+
+	if flags.sshCertKeyId == "" {
+		return fmt.Errorf("a ssh certificate key id is requiered")
```
tr1ck3r, line 705: "An ID is required for the SSH certificate (--id)"

cmd/vcert/validators.go
```diff
@@ -691,3 +693,86 @@ func validateSetPolicyFlags(commandName string) error {
...
+	}
+
+	err = validateExistingFile(flags.sshCertKeyId)
+	if err != nil {
+		return err
+	}
+
+	if flags.sshCertTemplate == "" {
+		return fmt.Errorf("ca value is required")
```
tr1ck3r, line 714: "CA value is required (--ca)"

cmd/vcert/commands.go
angelmoo: fixed on most recent commit

cmd/vcert/commands.go
angelmoo: fixed on most recent commit

cmd/vcert/validators.go
angelmoo: fixed on most recent commit.

cmd/vcert/commands.go
angelmoo: yes, and fixed in most recent commit

cmd/vcert/validators.go
angelmoo: as we changed the value to be template then I wrote something like this: "certificate issuing template value is required (--template)" please let me know if you agree with this or you have some changes on that.

cmd/vcert/flags.go
```diff
@@ -517,6 +529,101 @@ var (
...
+
+	flagSshCertGuid = &cli.StringFlag{
+		Name:        "guid",
+		Usage:       "A value that uniquely identifies the certificate request.",
+		Destination: &flags.sshCertGuid,
+	}
+
+	flagSshPrivKeyPassphrase = &cli.StringFlag{
+		Name: "pk-passphrase",
+		Usage: "The passphrase which will be used to wrap the generated private key before it is returned in the API response." +
```
tr1ck3r, line 612: Why can't we use `key-passphrase` for this use case?

cmd/vcert/flags.go
```diff
@@ -517,6 +529,101 @@ var (
...
+			" This is applicable only in case of service-generated keypair.",
+		Destination: &flags.sshCertGuid,
+	}
+
+	flagSshCertExtension = &cli.StringSliceFlag{
+		Name:  "extensions",
```
tr1ck3r, line 618: Should this be `extension` since you'd specify it multiple times if you need to add more than one extension?

tr1ck3r (approved):
Looks good @angelmoo I had a couple more questions.

cmd/vcert/flags.go
angelmoo: Changed on most recent commit.

cmd/vcert/flags.go
angelmoo: yes actually this is duplicated, key-passphrase is the one that we use

marcos-albornoz (approved):

ricrodriguezg (approved):

rvelaVenafi (approved):

Issue #183, opened 2021-10-28
vCert defaults to Server Certificate Type in TPP regardless of Certificate Template used


jmg421:
We believe vCert is somehow being hard coded to create a Server Certificate object in TPP.  The desired change would be to allow the Certificate type to be automatically detected by TPP (e.g. for Client Authentication Certificates/Templates)

visual-est:
@tr1ck3r I have tried the change and it works. However, I have not ran the tests :) I don't have environment.

PR #185, opened 2021-11-01
Allow TPP to autodetect the type of certificate objects

visual-est:
Potential fix for "vCert defaults to Server Certificate Type in TPP regardless of Certificate Template used" https://github.com/Venafi/vcert/issues/183

PS: I have tried the change and it works. However, I have not ran the tests :) I don't have environment.

### DIFF
--- a/pkg/venafi/tpp/connector.go
+++ b/pkg/venafi/tpp/connector.go
@@ -21,7 +21,6 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"fmt"
-	"github.com/Venafi/vcert/v4/pkg/policy"
 	"log"
 	"net/http"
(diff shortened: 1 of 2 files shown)


rvelaVenafi (approved):
Changes look good. We need to write a test suite to validate the correct certificate before release.

marcos-albornoz (approved):

achuchev:
@rvelaVenafi Good point. To be honest, I don't want to write tests :)

tr1ck3r:
We can take this without adding any tests because this behavior is entirely controlled by TPP and dependent upon the CA template assigned to the zone (policy folder).  It's something best covered by the tests for the `POST /vedsdk/certificates/request` method.

PR #188, opened 2021-11-17
Changes to handle SSH Protect 21.4

achuchev:
The PR contains changes which will make vCert comparable with SSH Protect 21.4 and prior. The high-level changes are:
- Ability to consume the issued certificates returned as response of the SSHCertificate/Request API method.
- Fix the logic which handles errors returned by SSH Protect.
- Ability to store CA public keys to file using '--file' flag.
- '--no-prompt' flag will be considered when asking to overwrite existing files (--file flag). When the flag is used, it will overwrite the existing files.
- '--id' (Key ID) is not required anymore as SSH Protect may have default value. This works for both SSH Protect 21.3 and 21.4
- //nolint added to some methods to prevent failed builds. This will be discussed with the maintainer.

### DIFF
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -67,17 +67,7 @@ jobs:
           asset_path: ./bin/linux/vcert
           asset_name: vcert_linux
           asset_content_type: application/octet-stream
-          
-      - name: Upload Darwin ARM artifact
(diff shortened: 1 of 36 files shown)


achuchev:
Due to commits made by mistake I am closing this PR. Soon a new will be opened.

PR #189, opened 2021-11-19
Changes to handle SSH Protect 21.4 (Updated)

achuchev:
The PR contains changes which will make vCert comparable with SSH Protect 21.4 and prior. The high-level changes are:

- Ability to consume the issued certificates returned as response of the SSHCertificate/Request API method.
- Fix the logic which handles errors returned by SSH Protect.
- Ability to store CA public keys to file using '--file' flag.
- '--no-prompt' flag will be considered when asking to overwrite existing files (--file flag). When the flag is used, it will overwrite the existing files.
- '--id' (Key ID) is not required anymore as SSH Protect may have default value. This works for both SSH Protect 21.3 and 21.4
/- /nolint added to some methods to prevent failed builds. This will be discussed with the maintainer.

PS: Apologues for the previous PR.

### DIFF
--- a/cmd/vcert/args.go
+++ b/cmd/vcert/args.go
@@ -128,5 +128,6 @@ type commandFlags struct {
 	sshCertExtension     stringSlice
 	sshCertPrincipal     stringSlice
 	sshCertWindows       bool
-	sshFile              string
+	sshFileCertEnroll    string
+	sshFileGetConfig     string
 }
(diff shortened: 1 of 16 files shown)


angelmoo (approved):
Changes looks good.

achuchev:
Thanks Angel. I will address the merge conflict soon.

luispresuelVenafi (approved):

PR #204, opened 2022-02-18
Implementation of methods to enable vssh tool

rvelaVenafi:
Added methods to allow the vssh tool to work properly with ssh certificates


ricrodriguezg (approved):

pkg/venafi/tpp/connector.go
```diff
@@ -270,6 +277,14 @@ func (c *Connector) RevokeAccessToken(auth *endpoint.Authentication) (err error)
 
 func processAuthData(c *Connector, url urlResource, data interface{}) (resp interface{}, err error) {
 
+	//isReachable, err := c.isAuthServerReachable()
```
achuchev, line 280: Hey @rvelaVenafi . Help me understand, what is the reason to have this commented?

I can provide why this method is needed even if Ping() is present. If you take a look on the URL of Ping() and this method you will see that they are using two different IIS application pools. This means that API clients will need bring them up in order to avoid timeouts. Let me know if you need additional details.

wallrj: The commented out code has since been deleted in https://github.com/Venafi/vcert/pull/435
 * https://github.com/Venafi/vcert/pull/435

The `isAuthServerReachable` function is never called AFAICS
 * https://github.com/search?q=repo%3AVenafi%2Fvcert%20isAuthServerReachable&type=code

PR #213, opened 2022-03-22
implement disabling subject fields.

angelmoo:
on policy specification now user can disable Subject fields by sending and empty array, i.e.: 
` "states": [ "" ] `


luispresuelVenafi (approved):
Changes looks good

ricrodriguezg (approved):

Issue #212, opened 2022-03-16
Support headless registration for Venafi as a Service


tr1ck3r:
_Venafi as a Service_ recently added a feature called "headless registration" that allows developers to  register for VaaS (and create a new company tenant if one doesn't already exist) without having to use a web browser and providing only their company email address.  VCert SDK should be updated to support this feature and VCert CLI should expose it through the existing `getcred` action.
```
vcert getcred [-u <url>] --email <email address> [--password <password>] [--format (text|json)]
```
If the CLI user specifies the new `--email` parameter then it is interpreted as a request to headlessly register for VaaS.  The output should display the inactive API key and tell the user to check their email inbox for an email with a link to activate (unless `--format json` is used in which case the output should include `apiKey` data from the API response; i.e. no messaging about the activation email).  The `--password` parameter should be optional mirroring the API call; when supplied, the specified password will allow the user to log into the VaaS web console assuming the user wasn't previously registered.  If a user has previously registered and their account has been configured to allow reset, headless registration can be used to rotate their API key (get a new one) which only becomes active if a link is clicked in an email triggered by the reset.  If the user's account does not allow reset an error is returned that should be output (true for any error).  As with other actions, the `--url` parameter overrides the https://api.venafi.cloud URL allowing for non-production testing by Venafi engineers. 
```
POST /v1/useraccounts
{
  "userAccountType": "API",
  "username": "person@company.com",
  "password": "myPassw0rd?"
}
```

Response includes unnecessary and mostly redundant data in "user" and "company" so those should simply be discarded.
```
HTTP 201 Created
{
    "user": {
        ...
    },
    "company": {
        ...
    },
    "apiKey": {
        "key": "aaaaaaaa-aaaa-aaaa-aaaaaaaaaaaaaaaaa",
        "userId": "bbbbbbbb-bbbb-bbbb-bbbbbbbbbbbbbbbbb",
        "username": "person@company.com",
        "companyId": "cccccccc-cccc-cccc-ccccccccccccccccc",
        "apiVersion": "ALL",
        "apiKeyStatus": "PENDING_ACTIVATION",
        "creationDate": "2022-03-16T15:40:38.533+00:00",
        "validityStartDate": "2022-03-16T15:40:38.533+00:00",
        "validityEndDate": "2022-09-12T15:40:38.533+00:00"
    }
}
```

marcos-albornoz:
Implemented on [v4.20.1](https://github.com/Venafi/vcert/tree/v4.20.1)

PR #219, opened 2022-05-11
Headless registration on VaaS

marcos-albornoz:


cmd/vcert/commands.go
```diff
@@ -613,100 +615,35 @@ func doCommandCredMgmt1(c *cli.Context) error {
...
-					fmt.Println("refresh_until: ", time.Unix(int64(resp.Refresh_until), 0).UTC().Format(time.RFC3339))
-				}
-			}
+	switch c.Command.Name {
+	case commandGetCredName:
+		if vaasConnector != nil {
+			return getVaaSCredentials(vaasConnector, &cfg)
 		} else {
-			return fmt.Errorf("Failed to determine credentials set")
+			return getTppCredentials(tppConnector, &cfg, clientP12)
```
rvelaVenafi, line 646: You should check for the instantiation of the tppConnector before running the command. What if there was an error instantiating the tppConnector?

cmd/vcert/commands.go
```diff
@@ -756,6 +693,137 @@ func doCommandCredMgmt1(c *cli.Context) error {
...
+
+		if flags.credFormat == "json" {
+			if err := outputJSON(apiKey); err != nil {
+				return err
+			}
+		} else {
+			var headerMessage string
+			if statusCode == http.StatusCreated {
+				headerMessage = "the user account was created successfully. To complete the registration please review your email account and follow the link."
+			} else { // it's assumed the statusCode is 202 which means the user exists and therefore the APIKey is rotated
```
rvelaVenafi, line 805: You should not assume the statusCode value and double check that it is indeed 202.

cmd/vcert/commands.go
```diff
@@ -756,6 +693,137 @@ func doCommandCredMgmt1(c *cli.Context) error {
...
+				headerMessage = "the user account already exists therefore the API Key was rotated. To complete the activation of the rotated API Key," +
+					" please review your email account and follow the link."
+			}
+
+			fmt.Println(headerMessage)
+			fmt.Println("key: ", apiKey.Key)
```
rvelaVenafi, line 811: Should all the response data be exposed in the cli? Is there any attribute that is not necessary to expose?

rvelaVenafi (changes requested):
There is no test suite for the new functionality. We need to have tests before we can commit these changes

cmd/vcert/commands.go
marcos-albornoz: the line 619 returns the connector if it was possible to instantiate otherwise an error is thrown, so if this part of the code is reached then we are sure that the connector was instantiate successfully.

cmd/vcert/commands.go
marcos-albornoz: The new function `parseUserDetailsResultFromPOST pkg/venafi/cloud/cloud.go line 384` which is the responsible to parse the response from VaaS only returns successfully if the statuscode is 201 or 202 so therefore it's redundant to do a validation to 202, but to avoid confusion on the understanding of the code logic maybe is a good idea to add the validation.

cmd/vcert/commands.go
marcos-albornoz: yes, that is the requirenment given in the issue https://github.com/Venafi/vcert/issues/212

cmd/vcert/commands.go
marcos-albornoz: Added the validation

luispresuelVenafi:
I agree with @rvelaVenafi , aside that, changes looks good

marcos-albornoz:
> I understand the whole process cannot be fully automated. However, we should test at least the API endpoint and make sure that part works. The creation of the emails should not be a problem, given we are not going to check the links we can create fake emails, unless VaaS has a way to verify an email exists indeed.

I've asked to Ryan about of this situation and he is agree that is not need any test for VCert side.

cmd/vcert/main.go
```diff
@@ -105,7 +105,7 @@ ACTIONS:
    getpolicy    To retrieve the certificate policy of a zone
    setpolicy    To apply a certificate policy specification to a zone
 
-   getcred      To obtain a new token for authentication
+   getcred      To obtain a new token for authentication on TPP or create a user account on VaaS
```
tr1ck3r, line 108: To obtain a new TPP authentication token or register a VaaS user account

marcos-albornoz: it's done

ricrodriguezg (approved):

luispresuelVenafi (approved):

rvelaVenafi (approved):

PR #235, opened 2022-06-17
Auto-create VaaS apps for headless registration

tr1ck3r:
VCert recently added support for VaaS headless registration but that user can't immediately begin requesting test certificates after they've registered because a VaaS application is required.  They could accomplish this by logging into the VaaS UI or use the `setpolicy` action but that is an undesirable extra step and the latter assumes they have administrative access to VaaS.  To simplify things and meet the objective of headless registration, VCert will be enhanced by this PR to automatically create VaaS applications if they are for the Default issuing template.  As such a new user would only need to invoke two actions to obtain a test certificate, `getcred` and `enroll`:
```
vcert getcred --email zach.jackson@venafi.example
vcert enroll -k xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx -z "Headless\Default" --cn test.venafi.example
```

### DIFF
--- a/pkg/venafi/cloud/connector.go
+++ b/pkg/venafi/cloud/connector.go
@@ -685,25 +685,29 @@ func (c *Connector) ReadPolicyConfiguration() (policy *endpoint.Policy, err erro
 // ReadZoneConfiguration reads the Zone information needed for generating and requesting a certificate from Venafi Cloud
 func (c *Connector) ReadZoneConfiguration() (config *endpoint.ZoneConfiguration, err error) {
 	var template *certificateTemplate
+	var statusCode int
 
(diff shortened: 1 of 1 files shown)


rvelaVenafi:
Would not be better to preemptively create the Application (or check if it exists) when the headless registration occurs?
So, the moment, the `getcred` operation returns you get both your apikey and your "default" Application.

`
apikey: XXXXXXXXXXXXXXXXX
application: Headless\Default
` 

Then you can invoke enroll using those parameters with no further changes into the core logic of vcert.

rvelaVenafi (approved):
Thanks for the explanation Ryan. This looks good

PR #238, opened 2022-06-30
Adding support for arm64 architecture on Linux

adillow:
This commit adds support for arm64 on Linux.  I can't install on AWS Graviton instances without this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ build_quick: get
 
 build: get
 	env GOOS=linux   GOARCH=amd64 go build $(GO_LDFLAGS) -o bin/linux/vcert         ./cmd/vcert
+	env GOOS=linux   GOARCH=arm64 go build $(GO_LDFLAGS) -o bin/linux/vcert_arm     ./cmd/vcert
 	env GOOS=linux   GOARCH=386   go build $(GO_LDFLAGS) -o bin/linux/vcert86       ./cmd/vcert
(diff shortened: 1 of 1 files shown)


adillow:
@angelmoo, @ricrodriguezg, @luispresuelVenafi, or @marcos-albornoz  would any of you be willing to review this?

adillow:
Changes are based on these PRs:

https://github.com/Venafi/vcert/pull/208
https://github.com/Venafi/vcert/pull/236

EduardoVV:
This looks like it was already addressed in another PR: https://github.com/Venafi/vcert/pull/218 and https://github.com/Venafi/vcert/pull/208

I'm closing it since intended functionality is already merged

luispresuelVenafi:
Hi @adillow , thank you for your contribution to the project!

Sorry for the delay into look into this, there have been some changes that we have made in order to properly sign VCert binaries so we took sometime before coming to look into support new OS/architecture. Could you pull recent change from master so it doesn't conflict? Take in count that there is also a change regarding the signature file, now it will be needed to be compressed inside the Linux zip file for the OS and architecture version as well as the binary. Once is done, we will also need to make a change in our end in order to support this new binary.

luispresuelVenafi:
This has been addressed in [v4.24.0](https://github.com/Venafi/vcert/releases/tag/v4.24.0)

Issue #199, opened 2022-01-14
Certificates/request is missing reenable


tehlers320:
**BUSINESS PROBLEM**

You can do this directly against the api (python)
```
re_enable_cert = {
                'Subject': myurl.foo.com,
                'PolicyDN': myfoler\\folder\folder\\,
                'Reenable': True
            }

re_enable_cert_req = requests.post(
        f"https://myvenafi/vedsdk/Certificates/Request",
        headers=headers,
        data=json.dumps(re_enable_cert),
        verify='/etc/ssl/certs/ca-certificates.crt'
    )
```

This is required to cleanup certs. 

**PROPOSED SOLUTION**
add the api call

Per the api docs see here on your local install `vedadmin/documentation/help/Content/SDK/WebSDK/r-SDK-POST-Certificates-request.htm`

```
Reenable | (Optional) The action to control a previously disabled certificate:false: Default. Do not renew a previously disabled certificate.true: Clear the Disabled attribute, reenable, and then renew the certificate (in this request). Reuse the same CertificateDN, that is also known as a Certificate object.
-- | --
```

**VENAFI EXPERIENCE**
using the api with python (without vcert lib) and trying to convert to go and using vcert.

luispresuelVenafi:
This have been addressed in [release v4.21.0](https://github.com/Venafi/vcert/releases/tag/v4.21.0)

PR #240, opened 2022-07-09
Set attribute Reenable=true to Request Cert

marcos-albornoz:
Fix for https://github.com/Venafi/vcert/issues/199
With this fix now the Certificates which are disabled will be
reenabled and renewed, this is the behavior expected to pass
the attribute "Reenable=true" into the request for
"POST Certificates/Request".


luispresuelVenafi (approved):
LGTM

EduardoVV (approved):

Issue #248, opened 2022-08-09
Add new methods for tpp connector


Geovanni89:
**BUSINESS PROBLEM**
vSSH command line utility is using now some logic to get the system version and user’s identity which belongs to vCert. Those methods should be consumed directly from vCert packages.

**PROPOSED SOLUTION**
We are planning to expose an existing endpoint on vCert to get the system version and create an internal method to get user’s identity and add the returned data on the tpp connector 

**CURRENT ALTERNATIVES**
Currently vSSH has a file with the necessary methods and vcert logic to work properly.

luispresuelVenafi:
This was addressed in VCert version [v4.22.1](https://github.com/Venafi/vcert/releases/tag/v4.22.1)

PR #249, opened 2022-08-09
Add user Identity  to TPP Connector and exposed the RetrieveSystemVersion() function

Geovanni89:
Related issue: https://github.com/Venafi/vcert/issues/248

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/Venafi/vcert/v4
 
 require (
+	github.com/blang/semver v3.5.1+incompatible
 	github.com/howeyc/gopass v0.0.0-20170109162249-bf9dde6d0d2c
 	github.com/pavel-v-chernykh/keystore-go/v4 v4.1.0
 	github.com/smartystreets/assertions v0.0.0-20180927180507-b2de0cb4f26d
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,6 @@
 module github.com/Venafi/vcert/v4
 
 require (
-	github.com/blang/semver v3.5.1+incompatible
 	github.com/howeyc/gopass v0.0.0-20170109162249-bf9dde6d0d2c
 	github.com/pavel-v-chernykh/keystore-go/v4 v4.1.0
 	github.com/smartystreets/assertions v0.0.0-20180927180507-b2de0cb4f26d
--- a/pkg/venafi/tpp/connector.go
+++ b/pkg/venafi/tpp/connector.go
@@ -437,14 +437,17 @@ func (c *Connector) RetrieveSelfIdentity() (response string, err error) {
 			return respIndentities.Identities[0].Name, nil
 		}
 	case http.StatusUnauthorized:
-		//return "", errs.NewExitAuthError(errors.New("authentication failure: check your credentials"))
 		return "", verror.AuthError
 	}
 	return "", fmt.Errorf("failed to get Self. Status code: %d, Status text: %s", statusCode, statusText)
 }
 
-// RetrieveServiceVersion returns the TPP system version of the connector context
 func (c *Connector) RetrieveServiceVersion() (string, error) {
+	return c.requestSystemVersion()
+}
+
+// requestSystemVersion returns the TPP system version of the connector context
+func (c *Connector) requestSystemVersion() (string, error) {
 	statusCode, status, body, err := c.request("GET", urlResourceSystemStatusVersion, "")
 	if err != nil {
 		return "", err
--- a/pkg/venafi/tpp/connector.go
+++ b/pkg/venafi/tpp/connector.go
@@ -442,12 +442,8 @@ func (c *Connector) RetrieveSelfIdentity() (response string, err error) {
 	return "", fmt.Errorf("failed to get Self. Status code: %d, Status text: %s", statusCode, statusText)
 }
 
-func (c *Connector) RetrieveServiceVersion() (string, error) {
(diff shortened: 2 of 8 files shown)


pkg/venafi/tpp/connector.go
```diff
@@ -414,8 +414,37 @@ func (c *Connector) requestMetadataItems(dn string) ([]metadataKeyValueSet, erro
 	return response.Data, err
 }
 
-// requestSystemVersion returns the TPP system version of the connector context
-func (c *Connector) requestSystemVersion() (string, error) {
```
marcos-albornoz, line 418: Please don't rename this function, instead you can create the function `RetrieveSefIdentity()` that you need and then call into this the` requestSystemVersion()` function.

marcos-albornoz (commented):
Can you provide unit tests for the new functions that you added?

pkg/venafi/tpp/connector.go
```diff
@@ -414,8 +414,37 @@ func (c *Connector) requestMetadataItems(dn string) ([]metadataKeyValueSet, erro
...
+		err = json.Unmarshal(body, respIndentities)
+		if err != nil {
+			return "", fmt.Errorf("failed to parse identity response: %s, body: %s", err, body)
+		}
+
+		if (respIndentities != nil) && (len(respIndentities.Identities) > 0) {
+			return respIndentities.Identities[0].Name, nil
+		}
+	case http.StatusUnauthorized:
+		//return "", errs.NewExitAuthError(errors.New("authentication failure: check your credentials"))
```
luispresuelVenafi, line 440: Could you remove commented code?

pkg/venafi/tpp/connector.go
```diff
@@ -414,6 +414,38 @@ func (c *Connector) requestMetadataItems(dn string) ([]metadataKeyValueSet, erro
...
+			return respIndentities.Identities[0].Name, nil
+		}
+	case http.StatusUnauthorized:
+		return "", verror.AuthError
+	}
+	return "", fmt.Errorf("failed to get Self. Status code: %d, Status text: %s", statusCode, statusText)
+}
+
+func (c *Connector) RetrieveServiceVersion() (string, error) {
+	return c.requestSystemVersion()
```
luispresuelVenafi, line 446: Why this method is being encapsulated? Won't be better to just rename `requestSystemVersion` to `RequestSystemVersion` so it is exposed outside the package in the connector interface (and by that, to avoid boilerplate code)?

Geovanni89: @luispresuelVenafi the encapsulation has been removed and the method was renamed instead

luispresuelVenafi (dismissed):
LGTM

go.mod
```diff
@@ -1,6 +1,7 @@
 module github.com/Venafi/vcert/v4
 
 require (
+	github.com/blang/semver v3.5.1+incompatible
```
marcos-albornoz, line 4: The latest version of this library is the 4.0.0 and it was released on May 2020 so this library is not updated since that time, so we need to evaluate if the benefit to use is worth the possible risks to use an old library.

pkg/venafi/tpp/connector_test.go
```diff
@@ -161,6 +162,67 @@ func TestAuthenticateAuthError(t *testing.T) {
...
+	serviceVersion, err := tpp.RetrieveSystemVersion()
+	if err != nil {
+		t.Fatalf("Failed to get Venafi service version. _bad_request_1_ Error: %v", err)
+	}
+
+	if serviceVersion == "" {
+		t.Fatalf("Failed to get Venafi service version. _bad_request_2_ Error: %v", err)
+	}
+
+	_, err = semver.Parse(strings.Join(strings.Split(serviceVersion, ".")[:3], "."))
```
marcos-albornoz, line 192: I think this can be replaced with other kind of validation like using Regex for that, and additionally this can allow us to avoid the dependency to the https://github.com/blang/semver library because only in this place is used

marcos-albornoz (changes requested):

pkg/venafi/tpp/connector_test.go
achuchev: @Geovanni89 Do you really need to parse the service version?

Geovanni89: @achuchev I don't, I'm removing that and using a regex validation instead. Also I'm removing the library that marcos mentioned above

go.mod
Geovanni89: @marcos-albornoz The library has been removed and a regex validation for the system version was added instead

marcos-albornoz (approved):

marcos-albornoz (approved):

luispresuelVenafi (approved):

achuchev:
@marcos-albornoz @luispresuelVenafi Thanks for working on this. Do you think we can merge this PR? It is blocking our work on the vSSH CLI project.

luispresuelVenafi:
Hi @Geovanni89 , by mistake we merged this branch and we discovered that it has issues since it broke our automated tests so we reverted the changes. I have let know @Geovanni89 about this. At the moment, as the changes are, we can't merge them to master.

luispresuelVenafi:
Hi @achuchev , by mistake we merged this branch and we discovered that it has issues since it broke our automated tests so we reverted the changes. I have let know @Geovanni89 about this. At the moment, as the changes are, we can't merge them to master.

PR #264, opened 2022-10-07
Consumers can now request private keys (for SSH certificates) protected by a passphrase

achuchev:
- The PrivateKeyPassphrase attribute was missing, so consumers cannot request private keys protected by a passphrase. 
- Remove PrivateKeyFormat from the SSH cert data structure. This option has never been implemented on the server side.

### DIFF
--- a/pkg/certificate/certificate.go
+++ b/pkg/certificate/certificate.go
@@ -227,7 +227,6 @@ type SshCertRequest struct {
 	Guid                      string
 	IncludePrivateKeyData     bool
 	PrivateKeyPassphrase      string
-	PrivateKeyFormat          string
 	IncludeCertificateDetails bool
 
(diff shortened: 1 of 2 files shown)


luispresuelVenafi:
Changes looks good @achuchev , could you add a unit test for this?

achuchev:
@luispresuelVenafi Do you mind to point me where I can find similar tests?

luispresuelVenafi:
Sure @achuchev, you could take this test as an example: https://github.com/Venafi/vcert/blob/master/pkg/venafi/tpp/connector_test.go#L2107

You could do the following tests:
- Provide the passphrase in the request and after the **retrieve** operation, that passphrase still works for opening the private key.
- Negative test: Provide a passphrase in the request and try to open the private key, it should fail.

Issue #239, opened 2022-07-06
Provide the ability to reset the certificate object in Venafi TPP


sitaramkm:
**BUSINESS PROBLEM**
If the downstream CA service is down for any reason, Venafi TPP changes the status of certificate object to Error.

The scenario to reproduce this is simple 
`cert-manager--->Venafi TPP--->MSCA`

`cert-manager` and `MSCA` could be replaced with any consumer and provider. 

- Request a certificate using `cert-manager`
- Stop the MSCA service 
- Trigger a certificate renewal using `cert-manager`
- As expected the renewal fails because Venafi cannot reach the downstream `MSCA` service 
- Renewal fails with a proper error that can be seen in the `CertificateRequest` resource 
- Venafi TPP during it's attempt to reach MSCA fails and marks the status of the certificate object as "Error" 
- Start the MSCA service
-  Trigger a manual renewal again. 
- This fails and there is no recovery. Unless the certificate object is `reset` any attempts to renew this certificate results in Error

**PROPOSED SOLUTION**
vCert provides a mechanism to reset the certificate object so consumers can attempt to heal the situation. 

**CURRENT ALTERNATIVES**
Currently, the only way to recover is to manually reset the certificate object in the UI and retry a renewal via API.

hawksight:
Perhaps this could be exposed in the `vcert` library, but also as a command line option as well?

E.g. a `reset` sub command something like:

```
USAGE:
   vcert [global options] command [command options] [arguments...]
   
ACTIONS:

   gencsr       To generate a certificate signing request (CSR)
   enroll       To enroll a certificate
   pickup       To retrieve a certificate
   renew        To renew a certificate
   reset          To reset a certificate state
   revoke       To revoke a certificate
```

Requiring either an `id` to target which certificate to reset:

```
vcert reset -u <HOST> -t <TOKEN> --id <CERT_ID> 
```

tr1ck3r:
Thank you for raising this issue @sitaramkm.  This problem is a side effect of TPP's object-based design and does not apply to Venafi as a Service.  As such, that gives me reservations about adding a new action (or a new option to an existing action) as @hawksight proposed.  Instead I think this "self-healing" should just be how VCert behaves.  No certificate request should ever be influenced by the success or failure of any previous certificate request.

### Guidance for anyone contributing this update to the project:
[...]

PR #269, opened 2022-11-08
Change RetrieveCertificate so that it resets the enrollment (if it exists) while retrieving the certificate

maelvls:
**Update 29 Nov 2022:** I put this PR back to "draft" because I am less and less confident with the changes I am making here. I propose to first agree on a set of `RetrieveCertificate` tests in #270 before I make any change to it `RetrieveCertificate`. When we have merged #270, we can proceed with the current PR.
**Update 5 Dec 2022:** I am now confident with the PR, it can now be reviewed.

---

**The problem:** When enrolling a new certificate, for example by running `vcert enroll` or when using cert-manager, people get "stuck" with an error of the like:

    500 Certificate \VED\Policy\Test\foo.com has encountered an error while processing,
    Status: This certificate cannot be processed while it is in an error state. Fix any
    errors, and then click Retry., Stage: 700.

This message occurs when a past enrollment has failed or an enrollment was still in progress for that certificate. The current workaround is to call to `POST /reset` with Restart=False, and then re-run the command `vcert enroll` (or renew the certificate in cert-manager).

This PR implements the design presented in [Solution 3: request, then reset if retrieve returns “Click Retry” or “WebSDK CertRequest”](https://hackmd.io/@maelvls/design-reset-cert-when-requesting#Solution-3-request-then-reset-if-retrieve-returns-%E2%80%9CClick-Retry%E2%80%9D-or-%E2%80%9CWebSDK-CertRequest%E2%80%9D). The OAuth token doesn't need to change (`certificate:manage`).

The [Jenkins build 2378](https://jenkins.eng.venafi.com/blue/organizations/jenkins/VCert%2Fvcert/detail/vcert/2378/pipeline) is passing for 63f7dff.

**Self-review:**

- I went with a "mock" HTTP server due to the difficulty to get the 202 and 500 HTTP status codes "on demand" from TPP. I didn't add a "live test" because I don't know how to consistently force TPP to return a 500 without RDP'ing into the VM and putting a PowerShell script or turning off the Windows CA (it can't be a policy check such as the domain, as this would be a "Stage 0" error which is the only stage number which gets reset upon requesting a new certificate).
  - **Fake server:** use a fake server to test it, but there is currently no fake server (note that @wallrj started to write a fake server in #262). The fake server, as opposed to the mock server, contains some logic.
  - **Mock responses:** use mock HTTP responses with a mock HTTP server.

**Manual test performed:**

Apart from the "mock" tests that I added, I also manually tested this change with a live TPP instance:

<details>
<summary>Reproducing the manual test</summary>

1. I spun up a TPP instance on CloudShare (using my Jetstack account), using the template "Master NGINX/K8s Blueprint" (snapshot "[6] training_2020-08-28") that comes with TPP 20.1.
2. I went to https://uvo1dq3vmkecfydcwxm.env.cloudshare.com/aperture/application-integrations/ and create a new API integration, since none of the existing contains `configuration:manage` (which I need to set the correct policy):
  
4. I created a token and stored it in the env var `TOKEN` with the following:
  
    ```sh
    TPP_URL=https://uvo1dq3vmkecfydcwxm.env.cloudshare.com
    TPP_USER=tppadmin
    TPP_PASSWORD=
    TOKEN=$(vcert getcred -u $TPP_URL --username $TPP_USER --password $TPP_PASSWORD --client-id=test --scope='certificate:manage;configuration:manage' --format json | jq -r .access_token)
    ```

5. Since we need to trigger an enrollment failure at a stage higher than 0, we will purposefully break the CA. To do that, I RDP'ed into the TPP VM, I opened the program `pkiview`, right-clicked on the only entry, "Manage CA...", and then clicked the "Stop" button:

    

6. I then enrolled:
  
    ```sh
    vcert enroll -u "$TPP_URL" -t "$TOKEN" --cn example.com -z 'TLS/SSL' --san-dns=example.com
    ```

    As expected, it shows:

    > vCert: 2022/11/14 14:58:38 unable to retrieve: Unexpected status code on TPP Certificate Retrieval. Status: 500 Certificate \VED\Policy\TLS/SSL\example.com has encountered an error while processing, Status: Post CSR failed with error: Cannot connect to the certificate authority (CA). Verify that your CA template settings are correct and that the remote server is available. For more information, search the Help system for Configuring the Microsoft Certificate Services Template Object., Stage: 500.

7. If we try to submit the certificate again, we still get the error at stage 500 (weirdly, the original disappeared and a generic message appears instead):
  
    > vCert: 2022/11/14 14:59:43 unable to retrieve: Unexpected status code on TPP Certificate Retrieval. Status: 500 Certificate \VED\Policy\TLS/SSL\example.com has encountered an error while processing, Status: This certificate cannot be processed while it is in an error state. Fix any errors, and then click Retry., Stage: 500.

8. I then fixed the CA:
  
    
   

9. I then tried to enroll again, and the "old" error with the sage 500 still shows:
  
    > vCert: 2022/11/14 15:06:30 unable to retrieve: Unexpected status code on TPP Certificate Retrieval. Status: 500 Certificate \VED\Policy\TLS/SSL\example.com has encountered an error while processing, Status: This certificate cannot be processed while it is in an error state. Fix any errors, and then click Retry., Stage: 500.

10. Now, from this project's folder, let us run `vcert` with the changes made in this PR:
    
    ```sh
    git fetch origin refs/pull/269/head:reset-before-request
    git checkout reset-before-request
    go run ./cmd/vcert enroll -u "$TPP_URL" -t "$TOKEN" --cn example.com -z 'TLS/SSL' --san-dns=example.com
    ```
  
    This time, the certificate gets properly requested.

</details>

**Changes in error messages:** this PR purposefully doesn't change any of the error messages.

**Possible breakages to other projects:**

- ✅  [terraform-provider-venafi](https://github.com/Venafi/terraform-provider-venafi) uses RetrieveCertificate in three locations. [One](https://github.com/Venafi/terraform-provider-venafi/blob/5374fa5ae86bb09609d228ef89f90a89261d3d3a/venafi/resource_venafi_certificate.go#L818) of them matches on the following specific error message:
    ```go
    "unable to retrieve: Unexpected status code on TPP Certificate Retrieval. Status: 400 Failed to lookup private key, error: Failed to lookup private key vault id"
     ```
     I added a test proving that this same error message will be shown in the same ciscumstances.
- ✅  [vault-pki-backend-venafi](https://github.com/Venafi/vault-pki-backend-venafi) won't be affected since it doesn't parses messages. It uses `RetrieveCertificate` [once](https://github.com/Venafi/vault-pki-backend-venafi/blob/88b14567c280d3d1b57addc2e51d4b5764e7b64e/plugin/pki/vcert_test.go#L61).
- ✅  [cert-manager](https://github.com/cert-manager/cert-manager) has a [single call](https://github.com/cert-manager/cert-manager/blob/d85e424cd0741bdb0fd4a5837d376c690e80fb9f/pkg/issuer/venafi/client/request.go#L65) to `RetrieveCertificate` and doesn't look at the messages.

### DIFF
--- a/pkg/venafi/tpp/connector.go
+++ b/pkg/venafi/tpp/connector.go
@@ -1232,9 +1232,32 @@ func (c *Connector) RetrieveCertificate(req *certificate.Request) (certificates 
 	}
 
 	startTime := time.Now()
+	retrieveCount := 0
 	for {
+		retrieveCount++
+
 		var retrieveResponse *certificateRetrieveResponse
 		retrieveResponse, err = c.retrieveCertificateOnce(certReq)
+
+		// As a workaround to certificates being stuck due to a past failed
+		// enrollment, we reset the certificate as part of RetrieveCertificate
+		// to avoid making an extra HTTP call when requesting a certificate. We
+		// only reset once, since it only makes sense to reset the past failed
+		// enrollment.
+		if shouldReset(err) && retrieveCount == 1 {
+			statusCode, status, _, err := c.request("POST", urlResourceCertificateReset, certificateResetRequest{
+				CertificateDN: req.PickupID,
+				Restart:       true,
+			})
+			if err != nil {
+				return nil, fmt.Errorf("while resetting certificate due to a past failed enrollment: %w", err)
+			}
+			if statusCode != http.StatusOK {
+				return nil, fmt.Errorf("while resetting certificate due to a past failed enrollment. Status: %s", status)
+			}
+
+			continue
+		}
 		if err != nil {
 			return nil, fmt.Errorf("unable to retrieve: %s", err)
 		}
@@ -1254,6 +1277,27 @@ func (c *Connector) RetrieveCertificate(req *certificate.Request) (certificates 
 		}
 		time.Sleep(2 * time.Second)
 	}
+}
+
+// After many tests, we found that the only way to know if the current
+// certificate request is stuck due to an old failed enrollment is to look for
+// "WebSDK CertRequest" or "Fix any errors, and then click Retry." in the
+// retrieve response when it returns a 500.
+func shouldReset(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	if !strings.Contains(err.Error(), "Status: 500") {
+		return false
+	}
+
+	if strings.Contains(err.Error(), "This certificate cannot be processed while it is in an error state. Fix any errors, and then click Retry.") ||
+		strings.Contains(err.Error(), "WebSDK CertRequest Module Requested Certificate") {
+		return true
+	}
+
+	return false
 }
 
 func (c *Connector) retrieveCertificateOnce(certReq certificateRetrieveRequest) (*certificateRetrieveResponse, error) {
--- a/pkg/venafi/tpp/connector_test.go
+++ b/pkg/venafi/tpp/connector_test.go
@@ -530,6 +530,48 @@ func TestRequestCertificateWithValidHours(t *testing.T) {
 	DoRequestCertificateWithValidHours(t, tpp)
 }
 
+func Test_shouldReset(t *testing.T) {
+	tests := []struct {
+		name     string
+		givenErr error
+		want     bool
+	}{
+		{
+			name:     "nil error",
+			givenErr: nil,
+			want:     false,
+		},
+		{
+			name:     "error is not a 500",
+			givenErr: fmt.Errorf("unable to retrieve: Unexpected status code on TPP Certificate Retrieval. Status: 400 Certificate does not exist."),
+			want:     false,
+		},
+		{
+			name:     "error is a 500 but not WebSDK or Click Retry",
+			givenErr: fmt.Errorf("unable to retrieve: Unexpected status code on TPP Certificate Retrieval. Status: 500 Certificate \\VED\\Policy\\TLS/SSL\\aexample.com has encountered an error while processing, Status: Post CSR failed with error: Cannot connect to the certificate authority (CA)., Stage: 500."),
+			want:     false,
+		},
+		{
+			name:     "error is a 500 and is Click Retry",
+			givenErr: fmt.Errorf("unable to retrieve: Unexpected status code on TPP Certificate Retrieval. Status: 500 Certificate \\VED\\Policy\\TLS/SSL\\aexample.com has encountered an error while processing, Status: This certificate cannot be processed while it is in an error state. Fix any errors, and then click Retry., Stage: 500."),
+			want:     true,
+		},
+		{
+			name:     "error is a 500 and is WebSDK",
+			givenErr: fmt.Errorf("unable to retrieve: Unexpected status code on TPP Certificate Retrieval. Status: 500 Certificate \\VED\\Policy\\TLS/SSL\\aexample.com has encountered an error while processing, Status: WebSDK CertRequest Module Requested Certificate, Stage: 500."),
+			want:     true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := shouldReset(tt.givenErr)
+			if got != tt.want {
+				t.Errorf("shouldReset() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
 // The reason we are using a mock HTTP server rather than the live TPP server is
 // because consistently triggering the 500 error in a stage different than 0
 // requires putting a powershell script on the TPP VM or turning the Microsoft
@@ -549,6 +591,7 @@ func TestRetrieveCertificate(t *testing.T) {
 	tests := []struct {
 		name         string
 		mockRetrieve []mockResp
+		mockReset    mockResp
 		givenTimeout time.Duration
 		expectErr    string
 	}{
@@ -587,6 +630,18 @@ func TestRetrieveCertificate(t *testing.T) {
 			expectErr: "unable to retrieve: Unexpected status code on TPP Certificate Retrieval. Status: 403 Failed to issue grant: User is not authorized for the requested scope",
 		},
 		{
+			name: "should fail when 'private key'",
+			mockRetrieve: []mockResp{
+				// This specific error message is relied upon by vcert itself as
+				// well as terraform-provider-venafi. So let's make sure we
+				// don't break it. See:
+				// https://github.com/Venafi/terraform-provider-venafi/blob/5374fa5/venafi/resource_venafi_certificate.go#L821
+				{"400 Failed to lookup private key, error: Failed to lookup private key vault id",
+					`{"Error":"Failed to lookup private key, error: Failed to lookup private key vault id"}`},
+			},
+			expectErr: "unable to retrieve: Unexpected status code on TPP Certificate Retrieval. Status: 400 Failed to lookup private key, error: Failed to lookup private key vault id",
+		},
+		{
 			name: "should succeed if cert immediately available regardless of the timeout value",
 			mockRetrieve: []mockResp{
 				{"200 OK",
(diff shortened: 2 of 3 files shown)


pkg/venafi/tpp/connector_test.go
```diff
@@ -51,29 +51,7 @@ var ctx *test.Context
 
 func init() {
 	ctx = test.GetEnvContext()
-	http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{
```
inteon, line 54: Why did you remove this code?

pkg/venafi/tpp/connector_test.go
```diff
@@ -528,6 +506,73 @@ func TestRequestCertificateWithValidHours(t *testing.T) {
...
+		switch {
+		case r.URL.Path == "/vedsdk/Identity/Self":
+			_, _ = w.Write([]byte("{}"))
+		case r.URL.Path == "/vedsdk/certificates/retrieve":
+			req := struct{ CertificateDN string }{}
+			_ = json.NewDecoder(r.Body).Decode(&req)
+			if req.CertificateDN != `\VED\Policy\Foo\test.com` {
+				t.Fatalf("/vedsdk/certificates/retrieve: expected CertificateDN to be %s but got %s", `\VED\Policy\Foo\test.com`, req.CertificateDN)
+			}
+			w.WriteHeader(500)
```
inteon, line 540: Would it be useful/ possible to also test a 202 case or a 404 case?

pkg/venafi/tpp/connector_test.go
maelvls: Oops. That is my mistake, I will re-add this bit.

pkg/venafi/tpp/connector_test.go
maelvls: Yes. I have added three other test cases: one with 200, one with 202, and one with an "unexpected" 500.

inteon: For context: the "unexpected" 500 is tested in the "should not reset when /certificates/retrieve returns an unrelated 500" function (I think)

maelvls: Yes. I'll make that clearer in the test name.

inteon (approved):
Thank you for fixing the remaining issues.

pkg/venafi/tpp/connector_test.go
```diff
@@ -528,6 +530,110 @@ func TestRequestCertificateWithValidHours(t *testing.T) {
...
+			givenResetStatus: 400,
+			givenResetBody:   `{"Error": "Reset is not completed. No reset is required for the certificate."}`,
+			expectRequest:    true,
+		},
+		{
+			name:             "should not request a certificate when reset returns an unexpected 401",
+			givenResetStatus: 401,
+			givenResetBody:   `{"error":"session_error","error_description":"Grant rejected scope 'certificate:manage'.}`,
+			expectRequest:    false,
+			expectErr:        "while resetting certificate in case of previous enrollment: 401 Unauthorized, body: ''",
```
maelvls, line 568: Note: I don't know why the body is `''`. I am expecting the body to be `{"error":"session_error","error_description":"Grant rejected scope 'certificate:manage'.}`, but it seems `request()` doesn't work like that...

maelvls: Ah, that was my fault. I fixed that in aad355f3d712fd870b5b29ead57441d2e949da60.

pkg/venafi/tpp/connector.go
```diff
@@ -678,19 +678,51 @@ func (c *Connector) proccessLocation(req *certificate.Request) error {
...
 	if req.Location != nil {
 		err = c.proccessLocation(req)
 		if err != nil {
 			return
 		}
 	}
+
+	// Let's unconditionally reset any prior pending or failed enrollment. This
+	// doesn't affect the existing certificate if one was already issued. At
+	// first, we wanted to decide whether to reset or not by first retrieving
```
JigarAtVenafi, line 696: Hi , I am not sure if I understand the rational behind reseting it every time.  This is edge case we are trying to solve here. 
I would expect Reset to only happen if certificate request is failing with certain error. This seems like an unnecessary API call every time when majority of time it wont be needed. 

Request Cert  --> If Failed with Error --> Reset --> Request again one more time. 
Request cert --> if Good --> Return

pkg/venafi/tpp/connector_test.go
```diff
@@ -28,13 +28,15 @@ import (
...
 	"net/http"
+	"net/http/httptest"
 	"net/url"
 	"os"
 	"path/filepath"
 	"reflect"
 	"regexp"
 	"strconv"
 	"strings"
+	"sync/atomic"
```
JigarAtVenafi, line 39: This need to be added in go dependencies. 
Jenkisn build is failing

`# github.com/Venafi/vcert/v4/pkg/venafi/tpp [github.com/Venafi/vcert/v4/pkg/venafi/tpp.test]

pkg/venafi/tpp/connector_test.go:579:116: undefined: atomic.Bool

pkg/venafi/tpp/connector_test.go:580:23: undefined: atomic.Bool

FAIL	github.com/Venafi/vcert/v4/pkg/venafi/tpp [build failed]

FAIL`

JigarAtVenafi (changes requested):

rvelaVenafi (changes requested):
There are too many unnecessary calls to reset if set unconditionally. 
The logic to reset based on the request response is better and not more complex than this PR. 
Just move this code to a separate function and validate the request response when statusCode == 500 and msg is the same as the expected JSON blob. 
You can even wrap the current logic into another function with a boolean parameter

```
func (c *Connector) RequestCertificate(req *certificate.Request) (requestID) (requestID string, err error) {
        // first try with reset
        id, err := RequestCertificateWithReset(req, true)
        if err != nil && err == CertResetError{
                 // Second try with no reset
                 id, err := RequestCertificateWithReset(req, false)
        }
}
```

Then have the actual code be moved to a second internal function:

```
func (c *Connector) RequestCertificateWithReset(req *certificate.Request, reset bool) (requestID) (requestID string, err error) {
    // Current RequestCertificate logic here
   
   	requestID, err = parseRequestResult(statusCode, status, body)
	if err != nil {
                 if reset {
		        //invoke reset logic here
                         err = resetCertificateResource(certDN, false)
                         if err != nil {
                                 // error that happened during reset 
                                 return "", err
                         }
                         // This error indicates the caller function that request failed because of 500 and a reset was attempted
                         return "", CertResetError       
                 }
                 // return normal error
                 return "", err 
	}
}
```

This way we can minimize the number of extra calls required for this use case

maelvls:
Thanks for the remarks. The choice of "resetting unconditionally" was motivated by (1) no benefit performance-wise when using the approach "reset only when failing", and (2) implementing "reset only when failing" is very tricky due to the number of undocumented corner cases. I documented the research that led to these results in the PR description.

> There are too many unnecessary calls to reset if set unconditionally.

What do you mean by unnecessary? This HTTP call doesn't seem to be costly (cf. the table with the duration in the PR description). Its cost is comparable to `POST /retrieve`.

pkg/venafi/tpp/connector.go
maelvls: I realize I haven't documented the reason why I am not first requesting, and then resetting if that fails.

The reason it is not possible is that `POST  /request` always succeeds as long as the given CSR or certificate parameters are valid. Calling `POST /request` doesn't allow you to know whether `POST /reset` needs to be called or not.

Does that make sense?

pkg/venafi/tpp/connector_test.go
maelvls: Hmm... `sync/atomic` comes from the Go standard library, which means it doesn't need to be referenced in `go.mod`. 

I will remove the use of `atomic.Bool`.

pkg/venafi/tpp/connector.go
maelvls: Since we settled on "Solution 3", I will resolve this conversation.

maelvls:
I finished implementing [Solution 3: request, then reset if retrieve returns “Click Retry” or “WebSDK CertRequest”](https://hackmd.io/@maelvls/design-reset-cert-when-requesting#Solution-3-request-then-reset-if-retrieve-returns-%E2%80%9CClick-Retry%E2%80%9D-or-%E2%80%9CWebSDK-CertRequest%E2%80%9D). Please take another look.

pkg/venafi/tpp/connector.go
```diff
@@ -1232,25 +1244,65 @@ func (c *Connector) RetrieveCertificate(req *certificate.Request) (certificates
...
+			// Continue below.
+		default:
+			return nil, fmt.Errorf("unexpected response while retrieving: %d: %s", statusCode, body)
 		}
+
 		if req.Timeout == 0 {
-			return nil, endpoint.ErrCertificatePending{CertificateID: req.PickupID, Status: retrieveResponse.Status}
+			return nil, endpoint.ErrCertificatePending{CertificateID: req.PickupID, Status: resp.Status}
 		}
 		if time.Now().After(startTime.Add(req.Timeout)) {
```
inteon, line 1304: Can you also fix this calculation and add the 2 seconds sleep period?
```suggestion
		if time.Now().Add(2 * time.Second).After(startTime.Add(req.Timeout)) {
```

pkg/venafi/tpp/connector.go
```diff
@@ -1232,25 +1244,65 @@ func (c *Connector) RetrieveCertificate(req *certificate.Request) (certificates
...
+				return nil, fmt.Errorf("while resetting certificate due to a past failed enrollment: %w", err)
+			}
+			if statusCode != http.StatusOK {
+				return nil, fmt.Errorf("while resetting certificate due to a past failed enrollment: %s, body: '%s'", status, body)
+			}
+
+			continue
+		case statusCode == http.StatusInternalServerError && resp.Status != "":
+			return nil, fmt.Errorf("enrollment failed at stage %d: %s", resp.Stage, resp.Status)
+		case statusCode == http.StatusOK:
```
inteon, line 1283: Instead of checking for `http.StatusOK` or `http.StatusAccepted`, can we just check `resp.CertificateData != ""` identical to the original code?

pkg/venafi/tpp/connector.go
```diff
@@ -1232,25 +1244,65 @@ func (c *Connector) RetrieveCertificate(req *certificate.Request) (certificates
...
+		switch {
+		// After many tests, we found that the only way to know if the current
+		// certificate request is stuck due to an old failed enrollment is to
+		// look for "WebSDK CertRequest" or "Fix any errors, and then click
+		// Retry." in the retrieve response. We only check this on the first
+		// retrieve because we know that these two messages always disappear
+		// after resetting.
+		case statusCode == http.StatusInternalServerError && retrieveCount == 1 &&
+			(resp.Status == "This certificate cannot be processed while it is in an error state. Fix any errors, and then click Retry." ||
+				resp.Status == "WebSDK CertRequest Module Requested Certificate"):
```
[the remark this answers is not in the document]
inteon: (this counter would have to be `==0` for a retry to happen)

pkg/venafi/tpp/connector.go
maelvls: I can replace `http.StatusOK` with `resp.CertificateData != ""`, but I would not be able to also do that for `http.StatusAccepted`, since I would fall into the `default` case (which results in an error).

pkg/venafi/tpp/connector.go
maelvls: I assumed that fixing this "bug" may break people, that's why I did not change it. But now that I think about it, I can't find any example of breakage that may happen if we fix this.

I'll do the change.

pkg/venafi/tpp/connector.go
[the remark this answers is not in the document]
maelvls: For context, Tim was getting that same message when one of Windows services was down, but that certificate wasn't in error.

@tr1ck3r In https://github.com/Venafi/vcert/issues/239#issuecomment-1242109588, you mentioned that TPP "will" return that body for a certificate object that was in error:

> It will have the following body after making an API request for a certificate object that was in error.
>
> ```json
> {
>   "Stage": 500, 
>   "Status": "WebSDK CertRequest Module Requested Certificate"
> }
> ```

Does this come from your knowledge of the TPP codebase, or is this something that you have observed over time?

pkg/venafi/tpp/connector.go
maelvls: Looking at the [`POST /retrieve` documentation](https://docs.venafi.com/Docs/current/TopNav/Content/SDK/WebSDK/r-SDK-POST-Certificates-retrieve.php), it seems like 200 is the only HTTP code for which `CertificateData` will be present. For 202, the only documented JSON fields are `Status` and `Stage`, `CertificateData` isn't meant to be returned on a 202.

But I'd prefer someone confirming that 202 never returns a `CertificateData`.

pkg/venafi/tpp/connector.go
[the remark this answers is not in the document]
maelvls: **tl;dr:** I'll go with the behavior we discussed, since "WebSDK CertRequest" means that there is a failed past enrollment in "sane" TPP instances.

**Context:** We had some back and forth with Tim. Since he has seen a case where TPP 22.04 would return the message "WebSDK CertRequest Module Requested Certificate" even though no prior enrollment existed, I wanted to investigate. Tim mentioned some Windows service was down that was causing the message. But I was not able to reproduce this issue. I assume it is a case of "misbehaving TPP" and won't happen in production TPP instances.

Tim suggested that we skip the "WebSDK CertRequest" messages until we find a "Click Retry" message.

I did a series of 100 tests with a TPP 20.1 server. Each time, I make sure the certificate is failing, then I request a certificate, and then I retrieve.

Case 1 occurs 60% of the time:

```
{"Stage":500,"Status":"This certificate cannot be processed while it is in an error state. Fix any errors, and then click Retry."}
{"Stage":500,"Status":"This certificate cannot be processed while it is in an error state. Fix any errors, and then click Retry."}
{"Stage":500,"Status":"This certificate cannot be processed while it is in an error state. Fix any errors, and then click Retry."}
{"Stage":500,"Status":"This certificate cannot be processed while it is in an error state. Fix any errors, and then click Retry."}
{"Stage":500,"Status":"This certificate cannot be processed while it is in an error state. Fix any errors, and then click Retry."}
```

Case 2 occurs 30% of the time:

```
{"Stage":500,"Status":"WebSDK CertRequest Module Requested Certificate"}
{"Stage":500,"Status":"This certificate cannot be processed while it is in an error state. Fix any errors, and then click Retry."}
{"Stage":500,"Status":"This certificate cannot be processed while it is in an error state. Fix any errors, and then click Retry."}
{"Stage":500,"Status":"This certificate cannot be processed while it is in an error state. Fix any errors, and then click Retry."}
{"Stage":500,"Status":"This certificate cannot be processed while it is in an error state. Fix any errors, and then click Retry."}
```

Case 3 occurs 10% of the time:

```
{"Stage":500,"Status":"WebSDK CertRequest Module Requested Certificate"}
{"Stage":500,"Status":"WebSDK CertRequest Module Requested Certificate"}
{"Stage":500,"Status":"WebSDK CertRequest Module Requested Certificate"}
{"Stage":500,"Status":"WebSDK CertRequest Module Requested Certificate"}
{"Stage":500,"Status":"WebSDK CertRequest Module Requested Certificate"}
```

Results:
- We cannot only solely on "Click Retry": we would miss 10% of the cases that we are trying to fix (cf. case 3).
- We cannot, in my mind, skip the "WebSDK CertRequest" messages until we see the "Click Retry" message, because it means we would miss 10% of the cases that we were trying to fix. Another minor problem with this approach is that we would end up waiting until the end of the timeout, and not reset.

pkg/venafi/tpp/tpp.go
```diff
@@ -624,25 +630,6 @@ func parseRequestData(b []byte) (data certificateRequestResponse, err error) {
 	return
 }
 
-func parseRetrieveResult(httpStatusCode int, httpStatus string, body []byte) (certificateRetrieveResponse, error) {
```
marcos-albornoz, line 627: Is it possible to keep these functions? that is with the idea to keep the approach to have the parsing of the results in dedicated functions

maelvls: I reverted my changes and made sure `parseRetrieveResult` stays. Please take another look. Thanks!

marcos-albornoz (approved):

luispresuelVenafi (approved):

pkg/venafi/tpp/connector.go
```diff
@@ -678,19 +678,51 @@ func (c *Connector) proccessLocation(req *certificate.Request) error {
...
+	// When the response to /reset is 400, we just ignore it; here are the three
+	// expected responses:
+	//
+	// | Happy responses                         | Description                |
+	// |-----------------------------------------|----------------------------|
+	// | 200 {"ProcessingResetCompleted": true}  | Success                    |
+	// | 400 {"Error":"CertificateDN error..."}  | Certificate DN not found   |
+	// | 400 {"Error":"No reset is required..."} | No enrollment to be reset  |
+	certDN := getCertificateDN(c.zone, req.Subject.CommonName)
+	statusCode, status, body, err := c.request("POST", urlResourceCertificateReset, struct {
```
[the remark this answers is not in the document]
maelvls: (from https://github.com/Venafi/vcert/pull/269#issuecomment-1321795062) It is not possible to use `POST /request` as a mean to know whether this certificate should be reset. The reason it is not possible is that `POST  /request` always succeeds as long as the given CSR or certificate parameters are valid. Calling `POST /request` doesn't allow you to know whether `POST /reset` needs to be called or not. I have elaborated a bit on this in https://github.com/Venafi/vcert/pull/269#discussion_r1027814433.

pkg/venafi/tpp/connector.go
```diff
@@ -1256,6 +1279,27 @@ func (c *Connector) RetrieveCertificate(req *certificate.Request) (certificates
...
+func shouldReset(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	if !strings.Contains(err.Error(), "Status: 500") {
+		return false
+	}
+
+	if strings.Contains(err.Error(), "This certificate cannot be processed while it is in an error state. Fix any errors, and then click Retry.") ||
```
JigarAtVenafi, line 1295: This is good. It is bit thorough checking. 
Just suggestion may be we can only search strings.Contains("error state") in first condition.  It might not be require if TPP will not be changing error.  No blocker.

maelvls: I agree, we could use a shorter substring to prevent potential future breakages if TPP changes this error message. I can change that in a later PR (this PR has being going for a while now).

rvelaVenafi (approved):

maelvls:
Announcement: this PR made it into vcert v4.23.0.

maelvls:
This improvement was reverted in 4.24.0. You can read https://github.com/Venafi/vcert/issues/273#issuecomment-1556938953 to learn more.

Issue #171, opened 2021-05-12
When retrieving a certificate with timeout the underlying error is hidden


barucoh:
**PROBLEM SUMMARY**
When retrieving a certificate with timeout set to anything other than 0 the returned error does not include the status code or the actual errors that occurred when trying to retrieve the certificate.
All we get is an "Operation timed out" and for all we know this could just mean a "slow network", but in the case there were at least more than 1 retries we need to know all the errors that occurred during every retry.
More specifically, every retry attempt we need to know the actual underlying error with its status code (if any).

This will greatly help troubleshooting errors

**STEPS TO REPRODUCE**
 - Create a certificate
 - Try to retrieve that certificate with `timeout > 0` (Make sure it is impossible to actually retrieve the certificate so the request to retrieve it fails)

**EXPECTED RESULTS**
Seeing a number of errors as the number of retries, each with it's status code

**ACTUAL RESULTS**
Just seeing a plain "Operation timed out" error

**ENVIRONMENT DETAILS**

**COMMENTS/WORKAROUNDS**

tr1ck3r:
Hi @barucoh, it sounds like the feature you may be seeking is provided by the VCert CLI's `--verbose` parameter.  When specified it is intended to show the individual REST API calls being made and their associated results.  For the following example, I make the request with the Venafi service stopped and then started it so there would be a few failed retrieval attempts:
```bash
vcert enroll -u https://tpp.venafi.example -t tn1PwE1QTZorXmvnTowSyA== -z DevOps --cn vcert-cli.venafi.example --san-dns vcert-cli.venafi.example --no-prompt --verbose`
```
```text
vCert: 2021/05/12 21:13:53 Successfully connected to TPP
vCert: 2021/05/12 21:13:53 Got 200 OK status for POST https://tpp.venafi.example/vedsdk/certificates/checkpolicy
vCert: 2021/05/12 21:13:53 Successfully read zone configuration for DevOps
vCert: 2021/05/12 21:13:53 Successfully created request for vcert-cli.venafi.example
vCert: 2021/05/12 21:13:54 Got 200 OK status for POST https://tpp.venafi.example/vedsdk/certificates/request
vCert: 2021/05/12 21:13:55 Got 200 OK status for POST https://tpp.venafi.example/vedsdk/metadata/get
vCert: 2021/05/12 21:13:55 Successfully posted request for vcert-cli.venafi.example, will pick up by \VED\Policy\DevOps\vcert-cli.venafi.example
[...]

PR #270, opened 2022-11-18
RetrieveCertificate: add mocked HTTP tests using real TPP responses

maelvls:
There are no tests that would allow us to "see" and discuss the issue reported in  https://github.com/Venafi/vcert/issues/171. Having tests, even statically mocked HTTP ones, would also allow us to prevent breakage (as much as possible) when changing RetrieveCertificate in #269.

I used `curl` and a 20.1 TPP VM to capture each individual HTTP response. At some point, I wish to document further how I produced these response samples.

The reason I went for a mocked HTTP server is because making the live TPP server return all of these messages is tricky; most 500 errors aren't easily reproducible unless we set up the e2e TPP VM to have e.g. one of its CA templates broken on purpose.
 
This "mocked HTTP server" approach isn't ideal because a drift may appear between earlier versions of TPP and these mocked responses, but I still think it is important to have them to prevent breakage when trying to refactor `RetrieveCertificate`.

### DIFF
--- a/pkg/venafi/tpp/connector_test.go
+++ b/pkg/venafi/tpp/connector_test.go
@@ -28,6 +28,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/http/httptest"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


marcos-albornoz (approved):

inteon (approved):
lgtm

luispresuelVenafi (approved):

inteon:
@luispresuelVenafi This PR should be ready to merge now.

PR #284, opened 2023-05-19
Fix regex domain name validation

inteon:
In TPP, we are used to specifying ".example.com" in the allowed domains, this restriction matches all subdomains of example.com, without matching example.com itself.
vCert does however not translate this rule into regex correctly, causing errors like the following:
```
Failed to request venafi certificate: common name test.example.com is not allowed in this policy: [^([\p{L}\p{N}-*]+\.)*\.example\.com$]
```

This PR fixes the regex generation logic and adds a test for this fix.

### DIFF
--- a/pkg/venafi/tpp/tpp.go
+++ b/pkg/venafi/tpp/tpp.go
@@ -808,6 +808,8 @@ func (sp serverPolicy) toZoneConfig(zc *endpoint.ZoneConfiguration) {
 }
 
 func (sp serverPolicy) toPolicy() (p endpoint.Policy) {
+	const allAllowedRegex = ".*"
+
 	addStartEnd := func(s string) string {
 		if !strings.HasPrefix(s, "^") {
 			s = "^" + s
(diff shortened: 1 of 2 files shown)


luispresuelVenafi:
Hi there @inteon 

Just tried this against master and worked fine:
```
$ go run main.go 
vCert: 2023/06/12 17:59:46 Successfully submitted certificate request. Will pickup certificate by ID \VED\Policy\path-to\test.example.com
vCert: 2023/06/12 17:59:49 Successfully picked up certificate for test.example.com
&{-----BEGIN CERTIFICATE-----
MIIIWjCCBkKgAwIBAgITbQC1Aq+I/ekHeXKaagAAALUCrzANBgkqhkiG9w0BAQsF
...
8akVTEDsmW7IrGArPDbqP8TD315fDXjCwHeplRUSmeWSCh3OZn9X/PC2Aq2nOBx6
//JPgYBkzdu1vzVf2Q7Xvf1nanNeY2PoMFmT0en8
-----END CERTIFICATE-----
```

This is what I have in the allowed_domains:
image

Which version of VCert and TPP you are using?

inteon:
> Hi there @inteon
> 
> Just tried this against master and worked fine:
> 
> ```
> $ go run main.go 
> vCert: 2023/06/12 17:59:46 Successfully submitted certificate request. Will pickup certificate by ID \VED\Policy\path-to\test.example.com
> vCert: 2023/06/12 17:59:49 Successfully picked up certificate for test.example.com
> &{-----BEGIN CERTIFICATE-----
> MIIIWjCCBkKgAwIBAgITbQC1Aq+I/ekHeXKaagAAALUCrzANBgkqhkiG9w0BAQsF
> ...
> 8akVTEDsmW7IrGArPDbqP8TD315fDXjCwHeplRUSmeWSCh3OZn9X/PC2Aq2nOBx6
> //JPgYBkzdu1vzVf2Q7Xvf1nanNeY2PoMFmT0en8
> -----END CERTIFICATE-----
> ```
> 
> This is what I have in the allowed_domains: image
> 
> Which version of VCert and TPP you are using? Also not sure if that's the what you meant to be in allowed domains. Does it look good to what you mention?

---

@luispresuelVenafi  This error only occurs when you run the local policy evaluation (which the VCert CLI does not run). This problem occurs when I use VCert as a Go library as shown below:
```go
zoneCfg, err := v.ReadZoneConfiguration(ctx)
if err != nil {
    return nil, err
}
zoneCfg.UpdateCertificateRequest(vreq)
err = zoneCfg.ValidateCertificateRequest(vreq)
if err != nil {
    return nil, err
}
```

luispresuelVenafi:
@inteon 

Got it, yes, I was able to reproduce the error as you mentioned. Tested code directly, changes looks good. Will involve more in team to get another approval.

luispresuelVenafi (approved):

marcos-albornoz (approved):

PR #293, opened 2023-06-21
NewRequest: set KeyCurve in NewRequest and add support for ed25519 for local CSR on VaaS

inteon:
Summary:

- The KeyCurve value was left empty in `NewRequest` for `ecdsa.PublicKey`s, this was marked as a TODO item in the code.
- This PR adds code to set the ED25519 `local` generated CSR support for VaaS
- Adds a test for this `NewRequest` function.
 - Added a switch case that sets KeyType for ed25519 keys.

EDIT: This PR does the above but support for ED25519 Keys are only available for local generate CSR's and, only supported by VaaS, so we have set up validation for any use related to TPP, since TPP doesn't support ED25519 keys for any of the supported methods. Also we have made necessary changes for this Keys, as they are using another type of algorithm different from ECDSA, there's a validation we do in:
`func (request *Request) CheckCertificate(certPEM string) error {`
that required us to treat ED25519 Keys as another different set of keys

**To be noted**, currently VaaS treats ED25519 Keys as another Key Curve for ECDSA, which is wrong but currently we are not doing change as of now, as it will impact how `Policy Management` is being used now.

### DIFF
--- a/pkg/endpoint/endpoint_test.go
+++ b/pkg/endpoint/endpoint_test.go
@@ -271,7 +271,7 @@ func TestBadKeyTypeValiateRequest(t *testing.T) {
 	}
 }
 
-func TestBadKeySizeValiateRequest(t *testing.T) {
+func TestBadKeySizeValidateRequest(t *testing.T) {
 	req := new(certificate.Request)
 	req.KeyType = certificate.KeyTypeRSA
 	req.KeyLength = 8192
@@ -288,7 +288,7 @@ func TestBadKeySizeValiateRequest(t *testing.T) {
 	}
 }
 
-func TestED25519KeyValiateRequest(t *testing.T) {
+func TestED25519KeyValidateRequest(t *testing.T) {
 	req := new(certificate.Request)
 	req.KeyType = certificate.KeyTypeED25519
 	req.KeyCurve = certificate.EllipticCurveED25519
@@ -297,9 +297,6 @@ func TestED25519KeyValiateRequest(t *testing.T) {
 	z.AllowedKeyConfigurations = []AllowedKeyConfiguration{
 		{
 			KeyType: certificate.KeyTypeED25519,
-			KeyCurves: []certificate.EllipticCurve{
-				certificate.EllipticCurveED25519,
-			},
 		},
 	}
 
--- a/pkg/endpoint/endpoint_test.go
+++ b/pkg/endpoint/endpoint_test.go
@@ -18,10 +18,11 @@ package endpoint
 
 import (
 	"crypto/x509"
-	"github.com/Venafi/vcert/v4/pkg/certificate"
 	"sort"
 	"strings"
 	"testing"
+
+	"github.com/Venafi/vcert/v4/pkg/certificate"
 )
 
 func TestNewZoneConfiguration(t *testing.T) {
@@ -153,6 +154,9 @@ func TestBadCNValiateRequest(t *testing.T) {
 	if err == nil {
 		t.Fatalf("CN should not have matched")
 	}
+	if !strings.HasSuffix(err.Error(), "common name vcert.test.bonjo.com is not allowed in this policy: [.*.vfidev.com .*.venafi.com]") {
+		t.Fatalf("Got unexpected error: %s", err)
+	}
 }
 
 func TestBadOValiateRequest(t *testing.T) {
@@ -165,6 +169,9 @@ func TestBadOValiateRequest(t *testing.T) {
 	err := z.ValidateCertificateRequest(req)
 	if err == nil {
 		t.Fatalf("O should not have matched")
+	}
+	if !strings.HasSuffix(err.Error(), "organization [Bonjo Org] doesn't match regular expressions: [Venafi.*]") {
+		t.Fatalf("Got unexpected error: %s", err)
 	}
 }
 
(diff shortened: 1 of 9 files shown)


luispresuelVenafi:
Hi @inteon 

I see that you added mocked tests for this, but currently we usually do real tests unless there's no way to easily do in pipeline. Since you are adding (enabling) a new EC key type, could you add a test for it as well? Like here we for EllipticCurveP521:
https://github.com/Venafi/vcert/blob/master/pkg/certificate/certificate_test.go#L167

inteon:
> Hi @inteon
> 
> I see that you added mocked tests for this, but currently we usually do real tests unless there's no way to easily do in pipeline. Since you are adding (enabling) a new EC key type, could you add a test for it as well? Like here we for EllipticCurveP521: https://github.com/Venafi/vcert/blob/master/pkg/certificate/certificate_test.go#L167

I added a test for EllipticCurveED25519.
I'm not 100% certain what you mean with "mocked tests". The tests that I added are indeed unit tests, this is because the functions that I'm testsing are mostly util/ helper functions.

luispresuelVenafi:
Sorry, you are right, it just that the way you described the test you did [here](https://github.com/Venafi/vcert/blob/b2d5d4dc0c5150087ee006b2b7b78a52847f02db/pkg/certificate/certificate_test.go#L562: on the quick review made me think that you were doing unit test with mocked data, but indeed you are doing real tests. But we don't need that block, do we? we are already testing those case with:
https://github.com/Venafi/vcert/blob/b2d5d4dc0c5150087ee006b2b7b78a52847f02db/pkg/certificate/certificate_test.go#L139
https://github.com/Venafi/vcert/blob/b2d5d4dc0c5150087ee006b2b7b78a52847f02db/pkg/certificate/certificate_test.go#L168
and most recently by your latest changes:
https://github.com/Venafi/vcert/blob/b2d5d4dc0c5150087ee006b2b7b78a52847f02db/pkg/certificate/certificate_test.go#L197

Could you remove this block then?
https://github.com/Venafi/vcert/blob/b2d5d4dc0c5150087ee006b2b7b78a52847f02db/pkg/certificate/certificate_test.go#L562-L614

inteon:
> Sorry, you are right, it just that the way you described the test you did [here](https://github.com/Venafi/vcert/blob/b2d5d4dc0c5150087ee006b2b7b78a52847f02db/pkg/certificate/certificate_test.go#L562): on the quick review made me think that you were doing unit test with mocked (faked) data, but indeed you are doing real tests. But we don't need that block, do we? we are already testing those cases with:
> 
> https://github.com/Venafi/vcert/blob/b2d5d4dc0c5150087ee006b2b7b78a52847f02db/pkg/certificate/certificate_test.go#L139
> 
> 
> https://github.com/Venafi/vcert/blob/b2d5d4dc0c5150087ee006b2b7b78a52847f02db/pkg/certificate/certificate_test.go#L168
> 
> 
> and most recently by your latest changes:
> https://github.com/Venafi/vcert/blob/b2d5d4dc0c5150087ee006b2b7b78a52847f02db/pkg/certificate/certificate_test.go#L197
> 
> Could you remove this block then?
> 
> https://github.com/Venafi/vcert/blob/b2d5d4dc0c5150087ee006b2b7b78a52847f02db/pkg/certificate/certificate_test.go#L562-L614

The only test that calls & tests `NewRequest` is `Test_NewRequest` as far as I can see. For that reason, I think it makes sense to keep that test.

luispresuelVenafi (dismissed):

pkg/venafi/cloud/cloud_test.go
```diff
@@ -86,7 +86,7 @@ func TestGenerateRequest(t *testing.T) {
 
 	keyTypeRSA := certificate.KeyTypeRSA
 	keyTypeEC := certificate.KeyTypeECDSA
-	keyTypeED25519 := certificate.KeyTypeED25519
+	KeyTypeED25519 := certificate.KeyTypeED25519
```
luispresuelVenafi, line 89: @inteon why did you change this to PascalCase?

inteon: thanks for noting this, this was a search-replace error.

pkg/endpoint/endpoint_test.go
```diff
@@ -153,6 +154,9 @@ func TestBadCNValiateRequest(t *testing.T) {
 	if err == nil {
 		t.Fatalf("CN should not have matched")
 	}
+	if !strings.HasSuffix(err.Error(), "common name vcert.test.bonjo.com is not allowed in this policy: [.*.vfidev.com .*.venafi.com]") {
```
inteon, line 157: This is an unrelated test improvement that makes sure that we get the expected error (before fixing getBaseZoneConfiguration, the returned error was actually the wrong error).

pkg/endpoint/endpoint_test.go
```diff
@@ -258,6 +283,30 @@ func TestBadKeySizeValiateRequest(t *testing.T) {
 	if err == nil {
 		t.Fatalf("Key size 8192 should not have been ok")
 	}
+	if !strings.HasSuffix(err.Error(), "the requested Key Type and Size do not match any of the allowed Key Types and Sizes") {
+		t.Fatalf("Got unexpected error: %s", err)
+	}
+}
+
+func TestED25519KeyValiateRequest(t *testing.T) {
```
luispresuelVenafi, line 291: Typo `TestED25519KeyValiateRequest`
->` TestED25519KeyValidateRequest`

rvelaVenafi (approved):

luispresuelVenafi (approved):

PR #297, opened 2023-07-05
Fix bug in IsCertNotFound function & add use function in test

inteon:
I missed this in https://github.com/Venafi/vcert/pull/295.
This PR fixes the IsCertNotFound function and ensures that it keeps working by using it in our test.

### DIFF
--- a/pkg/endpoint/endpoint.go
+++ b/pkg/endpoint/endpoint.go
@@ -89,7 +89,7 @@ type Connector interface {
 	GenerateRequest(config *ZoneConfiguration, req *certificate.Request) (err error)
 	// ResetCertificate resets the state of a TPP certificate.
 	// This function is idempotent, i.e., it won't fail if there is nothing to be reset.
-	// It returns an error of type *errCertNotFound if the certificate is not found.
+	// It returns an error of type *tpp.ErrCertNotFound if the certificate is not found.
 	ResetCertificate(req *certificate.Request, restart bool) (err error)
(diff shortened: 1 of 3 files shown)


luispresuelVenafi (approved):

pkg/endpoint/endpoint.go
```diff
@@ -90,7 +90,7 @@ type Connector interface {
 	GenerateRequest(config *ZoneConfiguration, req *certificate.Request) (err error)
 	// ResetCertificate resets the state of a TPP certificate.
 	// This function is idempotent, i.e., it won't fail if there is nothing to be reset.
-	// It returns an error of type *errCertNotFound if the certificate is not found.
+	// It returns an error of type *tpp.ErrCertNotFound if the certificate is not found.
```
marcos-albornoz, line 93: This change in the doc is not correct given this is the endpoint interface which is generic and indepent of the platform which VCert is connecting and then we can't refer here to specific elements to tpp platform

luispresuelVenafi: @marcos-albornoz Yes, currently `ResetCertificate` is not supported for VaaS. It's TPP only (also, that structure only lives on TPP's side of things for now):
https://github.com/Venafi/vcert/blob/master/pkg/venafi/cloud/connector.go#L843-L845

luispresuelVenafi: Resolved by latest changes

rvelaVenafi (approved):

PR #298, opened 2023-07-05
Fix TODO: Add validation for email, ip and uris

inteon:
This PR fixes a TODO in the VCert code.

### DIFF
--- a/pkg/endpoint/endpoint.go
+++ b/pkg/endpoint/endpoint.go
@@ -255,9 +255,11 @@ func (p *Policy) ValidateCertificateRequest(request *certificate.Request) error 
 		if err != nil {
 			return err
 		}
+
 		if !isComponentValid(parsedCSR.EmailAddresses, p.EmailSanRegExs, true) {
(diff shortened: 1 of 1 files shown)


pkg/endpoint/endpoint.go
```diff
@@ -280,17 +284,18 @@ func (p *Policy) ValidateCertificateRequest(request *certificate.Request) error
 			return fmt.Errorf(organizationUnitError, parsedCSR.Subject.OrganizationalUnit, p.SubjectOURegexes)
 		}
 
-		if !isComponentValid(parsedCSR.Subject.Country, p.SubjectCRegexes, false) {
-			return fmt.Errorf(countryError, parsedCSR.Subject.Country, p.SubjectCRegexes)
+		if !isComponentValid(parsedCSR.Subject.Province, p.SubjectSTRegexes, false) {
```
inteon, line 287: I changed the order, so it matches the ordering that is used below. This should not have any noteworthy effects on the behavior of this function.

PR #333, opened 2023-08-07
Feat: Add enrollment support for Firefly

marcos-albornoz:
This PR contains the following changes:
- flags and variables related to the "tppToken" concept were renamed to "token" given it will be used also to provide the Firefly access token.
- added valid-period flag to support validity periods in ISO 8601 format for Firefly.
- added support to use valid-days flag instead of valid-period for Firefly.

Closes https://venafi.atlassian.net/browse/VC-24605, https://venafi.atlassian.net/browse/VC-24606, https://venafi.atlassian.net/browse/VC-24607 and https://venafi.atlassian.net/browse/VC-24608


rvelaVenafi (approved):

cmd/vcert/args.go
```diff
@@ -113,6 +113,7 @@ type commandFlags struct {
 	csrFormat            string
 	credFormat           string
 	validDays            string
+	validPeriod          string
```
luispresuelVenafi, line 116: @marcos-albornoz could you remove logic for using `validPeriod` in favor of using `validDays` and/or `validDuration` (this depends if talking about SDK or CLI)? You can base about ISO 8601 format translation is handled in VaaS for both SDK and CLI cases (other platform that also handles ISO 8601 format).

luispresuelVenafi (changes requested):

cmd/vcert/args.go
marcos-albornoz: It´s not needed to remove it, because the support for Valid Days is still available for Firefly. This flag is a new Flag and is to provide support for a valid Period in ISO-8601 format.

luispresuelVenafi: My comment was that I don't see use, and probably will create more complexity, to the user that we have two flags for handling the same final end, giving a validity to the certificate. IMHO, introducing this new flag go out of scope of supporting Firefly, but as we discussed in DM, I'll accept this changes while we don't release it yet until we discuss this in further presentation of VCert.

luispresuelVenafi (approved):

PR #341, opened 2023-08-14
vcert/v4/pkg: Expose API for refreshing the access token in Connector interface

KalpitTiwari03:
"RefreshAccessToken()" api is not exposed in Connector interface at the moment. Adding the same by introducing a wrapper over "RefreshAccessToken()" which returns an interface instead of TPP specific struct.

Other possible approach can be - moving out the struct "OauthRefreshAccessTokenResponse" to endpoint.go file and add
"RefreshAccessToken()" to Connector interface. But the struct is TPP specific so placing it in endpoint pkg does not look right therefor going with current approach.


KalpitTiwari03:
Could I please get a review here?

PR #343, opened 2023-08-14
Feat: Add support for Device Flow grant for Firefly

marcos-albornoz:
This PR provides the support for OAuth 2.0 Device Flow Grant:

- Added the `device-url` flag to provide the device url to request a device code.

Closes https://venafi.atlassian.net/browse/VC-24604


pkg/venafi/firefly/connector.go
```diff
@@ -114,6 +114,11 @@ func (c *Connector) Authorize(auth *endpoint.Authentication) (token *oauth2.Toke
 		return config.PasswordCredentialsToken(context.Background(), auth.User, auth.Password)
 	}
 
+	// if it's a device flow grant
+	if auth.IdentityProvider.DeviceURL != "" {
+		return c.getDeviceAccessToken(auth)
+	}
+
 	// if it's a client credentials flow grant
 	if auth.ClientSecret != "" {
```
rvelaVenafi, line 123: ClientSecret should be the first authorization method we check. As user/password usage is discouraged and device is a niche use case.

pkg/venafi/firefly/deviceFlow.go
```diff
@@ -0,0 +1,197 @@
...
+	//setting as default verificationURL the value returned in VerificationURI given that is the used for the most of the
+	// OAuth IdP like Okta, Auth0 and WSO2 return the verification_uri to show to the user
+	verificationURL := devCred.VerificationURI
+	//if that is empty then trying to use the VerificationURL given google uses verification_url to
+	//return the verification_uri to show to the user
+	if verificationURL == "" {
+		verificationURL = devCred.VerificationURL
+	}
+
+	fmt.Printf("please go to %v and enter the code %v . It will expire in %dm and %ds\n\n", verificationURL, devCred.UserCode, devCred.ExpiresIn/60, devCred.ExpiresIn%60)
```
rvelaVenafi, line 60: We should make a bigger and better message here as this is really important the user see it.
One example is how aws-cli does it.

pkg/venafi/firefly/deviceFlow.go
```diff
@@ -0,0 +1,197 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		//parsing the response
+		token, err := parseWaitingDeviceAuthorizationRequestResult(statusCode, body)
+
+		if err != nil {
+			switch err.Error() {
+			case "authorization_pending":
```
rvelaVenafi, line 140: Make this strings constants

pkg/venafi/firefly/deviceFlow.go
```diff
@@ -0,0 +1,197 @@
...
+				devCred.Interval += 5
+				time.Sleep(time.Duration(devCred.Interval) * time.Second)
+			case "access_denied":
+				return nil, fmt.Errorf("the access from device was denied by the user")
+			case "expired_token":
+				return nil, fmt.Errorf("the device code expired")
+			default:
+				return nil, fmt.Errorf("the authorization failed. %w", err)
+			}
+		} else {
```
rvelaVenafi, line 152: invert this if/else to remove nesting. If err == nil return with token. Then no else is necessary and you can follow the next steps to check the error and the switch.

pkg/venafi/firefly/firefly.go
```diff
@@ -90,20 +92,33 @@ func (c *Connector) GenerateRequest(_ *endpoint.ZoneConfiguration, req *certific
...
 	var payload io.Reader
 	var b []byte
+	var values url.Values
+
+	contentType := "application/json"
+
 	if method == "POST" || method == "PUT" {
-		b, _ = json.Marshal(data)
-		payload = bytes.NewReader(b)
+		//determining if the data is type of url.Values
```
rvelaVenafi, line 103: Do this in a different function to keep code easy to read

rvelaVenafi (changes requested):

pkg/venafi/firefly/connector.go
luispresuelVenafi: Level priority:
- ClientSecret
- Device
- User/Password

pkg/venafi/firefly/connector.go
```diff
@@ -114,6 +114,11 @@ func (c *Connector) Authorize(auth *endpoint.Authentication) (token *oauth2.Toke
 		return config.PasswordCredentialsToken(context.Background(), auth.User, auth.Password)
 	}
 
+	// if it's a device flow grant
+	if auth.IdentityProvider.DeviceURL != "" {
+		return c.getDeviceAccessToken(auth)
```
luispresuelVenafi, line 119: You can seperate the function from here and verify err, if not error then print log success log and return values:
```
if auth.IdentityProvider.DeviceURL != "" {
                authToken, err = c.getDeviceAccessToken(auth)
                if err != nil {
                   return nil, err
                }
                log.Printf("Successfully authorized device) // Not sure about the log syntax
		return authToken, nil
}
```

pkg/venafi/firefly/connector.go
marcos-albornoz: done

pkg/venafi/firefly/deviceFlow.go
marcos-albornoz: done

pkg/venafi/firefly/deviceFlow.go
marcos-albornoz: done

pkg/venafi/firefly/deviceFlow.go
marcos-albornoz: done

pkg/venafi/firefly/connector.go
marcos-albornoz: It's done now into the c.getDeviceAccessToken(auth)

pkg/venafi/firefly/firefly.go
marcos-albornoz: I prefer to keep in this way given here are gotten the payload and the values which only has meaning when the request data is of type url.Values which is expected when the request is for device flow grant given the request is needed to do using form params and these values are stored, if something is wrong, to be used in the build of the error.

rvelaVenafi (approved):

luispresuelVenafi (approved):

PR #350, opened 2023-08-16
fix: Normalize Firefly url

luispresuelVenafi:


pkg/venafi/firefly/connector.go
```diff
@@ -59,7 +60,18 @@ func (c *Connector) RetrieveAvailableSSHTemplates() (response []certificate.SshA
...
 func NewConnector(url string, zone string, verbose bool, trust *x509.CertPool) (*Connector, error) {
-	return &Connector{baseURL: url, zone: zone, verbose: verbose, trust: trust}, nil
+	normalizedUrl, err := normalizeURL(url)
+	if err != nil {
+		return nil, fmt.Errorf("%w: failed to normalize URL: %v", verror.UserDataError, err)
+	}
+	return &Connector{baseURL: normalizedUrl, zone: zone, verbose: verbose, trust: trust}, nil
+}
+
+func normalizeURL(url string) (normalizedURL string, err error) {
```
EduardoVV, line 70: Is this extra function needed?

luispresuelVenafi: Good catch

luispresuelVenafi: @EduardoVV fixed by latest commit

EduardoVV (approved):

pkg/venafi/cloud/connector.go
```diff
@@ -673,16 +672,9 @@ func normalizeURL(url string) (normalizedURL string, err error) {
 		url = apiURL
 		//return "", fmt.Errorf("base URL cannot be empty")
```
rvelaVenafi, line 673: Remove this comment as no longer needed or return right away as no further modifications are required to the default URL

pkg/venafi/cloud/connector.go
```diff
@@ -673,16 +672,9 @@ func normalizeURL(url string) (normalizedURL string, err error) {
...
-	} else {
-		modified = reg.ReplaceAllString(modified, "https://")
-	}
-	if !strings.HasSuffix(modified, "/") {
-		modified = modified + "/"
-	}
+
+	modified := util.NormalizeUrl(url)
+
 	normalizedURL = modified
```
rvelaVenafi, line 678: This step can be removed, there is no further modification to the url after returning from `util.NormalizeUrl()` function

pkg/venafi/firefly/connector.go
```diff
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Venafi/vcert/v5/pkg/util"
```
rvelaVenafi, line 28: Sort this import accordingly to our rules

pkg/venafi/firefly/connector.go
```diff
@@ -59,7 +60,8 @@ func (c *Connector) RetrieveAvailableSSHTemplates() (response []certificate.SshA
 
 // NewConnector creates a new Firefly Connector object used to communicate with Firefly
 func NewConnector(url string, zone string, verbose bool, trust *x509.CertPool) (*Connector, error) {
-	return &Connector{baseURL: url, zone: zone, verbose: verbose, trust: trust}, nil
+	normalizedUrl := util.NormalizeUrl(url)
```
rvelaVenafi, line 63: Move this line to a NormalizeUrl function like in other connectors and add a validation for empty url. Just like other connectors, make the NormalizeUrl return an error and use it here in the constructor.

rvelaVenafi:
Also, I noticed NewConnector() in cloud connector just returns the error when normalizeUrl fails. While TPP wraps the error. We should standardize and use the same error as TPP is using for Cloud and Firefly connectors in their constructors.

rvelaVenafi (changes requested):

pkg/venafi/firefly/connector.go
marcos-albornoz: It's done

pkg/venafi/cloud/connector.go
marcos-albornoz: It's done

pkg/venafi/cloud/connector.go
marcos-albornoz: It's done

pkg/venafi/firefly/connector.go
marcos-albornoz: It's done

pkg/venafi/firefly/connector.go
```diff
@@ -59,9 +60,22 @@ func (c *Connector) RetrieveAvailableSSHTemplates() (response []certificate.SshA
 
 // NewConnector creates a new Firefly Connector object used to communicate with Firefly
 func NewConnector(url string, zone string, verbose bool, trust *x509.CertPool) (*Connector, error) {
+	if url != "" {
+		var err error
```
rvelaVenafi, line 64: This declaration is not required.
below line can do
`url, err := normalizeUrl()`

rvelaVenafi (changes requested):

pkg/venafi/firefly/connector.go
marcos-albornoz: that is not correct, given that create a new `user` var which only lives in the scope of the surrounder `if condition`

rvelaVenafi: got it

rvelaVenafi (approved):

PR #381, opened 2023-08-28
Feat: Fix support for default key-curve

marcos-albornoz:
Fix that the default value for `--key-type ecdsa` must be `P256`

### DIFF
--- a/pkg/venafi/firefly/connector.go
+++ b/pkg/venafi/firefly/connector.go
@@ -313,11 +313,11 @@ func (c *Connector) getCertificateRequest(req *certificate.Request) (*certificat
 	keyAlgorithm := ""
 	switch req.KeyType {
 	case certificate.KeyTypeRSA:
-		keyLength, err := GetRSASize(req.KeyLength)
+		keySize, err := GetRSASize(req.KeyLength)
 		if err != nil {
 			return nil, err
 		}
-		keyAlgorithm = fmt.Sprintf("RSA_%d", keyLength)
+		keyAlgorithm = fmt.Sprintf("RSA_%d", keySize)
 	case certificate.KeyTypeECDSA, certificate.KeyTypeED25519:
 		keyCurve := req.KeyCurve
 		if keyCurve == certificate.EllipticCurveNotSet {
--- a/pkg/venafi/firefly/connector.go
+++ b/pkg/venafi/firefly/connector.go
@@ -200,7 +200,11 @@ func (c *Connector) SynchronousRequestCertificate(req *certificate.Request) (cer
 
 	zap.L().Info("requesting certificate", zap.String("cn", req.Subject.CommonName), fieldPlatform)
 	//creating the request object
-	certReq := c.getCertificateRequest(req)
+	certReq, err := c.getCertificateRequest(req)
+	if err != nil {
+		zap.L().Error("HTTP request failed", fieldPlatform, zap.Error(err))
+		return nil, err
+	}
 
 	zap.L().Info("sending HTTP request", fieldPlatform)
 	statusCode, status, body, err := c.request("POST", c.getCertificateRequestUrl(req), certReq)
@@ -228,7 +232,7 @@ func (c *Connector) SynchronousRequestCertificate(req *certificate.Request) (cer
 	return certificates, nil
 }
 
-func (c *Connector) getCertificateRequest(req *certificate.Request) *certificateRequest {
+func (c *Connector) getCertificateRequest(req *certificate.Request) (*certificateRequest, error) {
 	zap.L().Info("building certificate request", fieldPlatform)
 	fireflyCertRequest := &certificateRequest{}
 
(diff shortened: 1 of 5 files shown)


EduardoVV (approved):

pkg/venafi/firefly/connector.go
```diff
@@ -309,15 +313,22 @@ func (c *Connector) getCertificateRequest(req *certificate.Request) *certificate
 	keyAlgorithm := ""
 	switch req.KeyType {
 	case certificate.KeyTypeRSA:
-		keyAlgorithm = fmt.Sprintf("RSA_%d", req.KeyLength)
-
+		keyLength, err := GetRSASize(req.KeyLength)
```
rvelaVenafi, line 316: change the var name to `keySize` instead of `keyLength`

pkg/venafi/firefly/connector.go
```diff
@@ -228,7 +232,7 @@ func (c *Connector) SynchronousRequestCertificate(req *certificate.Request) (cer
 	return certificates, nil
 }
 
-func (c *Connector) getCertificateRequest(req *certificate.Request) *certificateRequest {
+func (c *Connector) getCertificateRequest(req *certificate.Request) (*certificateRequest, error) {
```
rvelaVenafi, line 235: at what point an error is returned by this function? No need to return an error if you never throw one

pkg/venafi/firefly/fireflyUtil.go
```diff
@@ -0,0 +1,24 @@
...
+	if rsaSizeResult == 0 {
+		rsaSizeResult = certificate.DefaultRSAlength
+	}
+
+	if !rsaSizes[rsaSize] {
+		var sizes []int
+		for size, _ := range rsaSizes {
+			sizes = append(sizes, size)
+		}
+		return 0, fmt.Errorf("key Size %d is not supported. The valid RSA sizes for Firefly are %v", rsaSize, sizes)
```
rvelaVenafi, line 20: Remove `The` word after the dot:
`key Size %d is not supported. Valid RSA sizes for Firefly are %v`

pkg/venafi/firefly/firefly.go
```diff
@@ -43,6 +43,10 @@ const (
 	scopesSeparator = " "
 )
 
+var (
+	rsaSizes = map[int]bool{certificate.DefaultRSAlength: true, 3072: true, 4096: true}
```
rvelaVenafi, line 47: Cant this be achieved with a simple array. Values are finite and array is small (4 or 5 values max). Iterating over such array is not expensive an easier to read and print on the error message

rvelaVenafi (changes requested):

pkg/venafi/firefly/connector.go
marcos-albornoz: keyLength is the concept used by CertificateRequest structure so I've decided to keep the var on that way

pkg/venafi/firefly/connector.go
marcos-albornoz: The error commented in the below comment it's the one can be thrown

pkg/venafi/firefly/connector.go
rvelaVenafi: then change the name of the function. size is the correct term to refer to the RSA key type. However if we are going to use length, then lets use it consistently everywhere

pkg/venafi/firefly/firefly.go
marcos-albornoz: I've implemented a kind of set as it's explained in the official documentation https://go.dev/doc/effective_go#maps

pkg/venafi/firefly/fireflyUtil.go
marcos-albornoz: Done

pkg/venafi/firefly/connector.go
marcos-albornoz: Agree...

pkg/venafi/firefly/connector.go
rvelaVenafi: Ok, I totally lost grasp of the scope

pkg/venafi/firefly/connector.go
marcos-albornoz: Done

pkg/venafi/firefly/firefly.go
rvelaVenafi: I get it. However given the limited values on the array and the small length, this could be written just by using a simple array and comparing values. 

In the end you are not saving any performance or loop because you are already looping the entire map in order to build the values on the error message. I think it would be easier to read just to iterate the values array to find the size and jus print the array in the message as is. Its one loop in both approaches but:

- By using a map, you always need to loop the maximum number of iterations to build the error message.
- It is harder to read and understand what is happening.

marcos-albornoz: The iteration you mention only will happen when the passed rsa-size is not a valid one in order to be used for the error to thrown, when with the approach you mention the iteration will occur no matter if the rsa-size value is valid or not.

rvelaVenafi (approved):

PR #386, opened 2023-09-04
feat(playbook): Added attributes to CAPI installation

rvelaVenafi:
- Added capiFriendlyName field to installation object 
- Added capiLocation field to installation object
- Deprecated location field from installation object
- Added warning when using deprecated location field 
- Added warning when capiFriendlyName is not set

- Now CAPI installations will use the value set in `capiLocation`. If not set, then will use the value from `location`
- CAPI installations will use the value set in `capiFriendlyName`. If not set, then will use the value from the `request.subject.commonName` (Check) or `certificate.subject.commonName` (install)
- A random password will be generated for the P12 bundle used to import the certificate into CAPI store every time a certificate is installed

### DIFF
--- a/pkg/playbook/app/installer/capi.go
+++ b/pkg/playbook/app/installer/capi.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/Venafi/vcert/v5/pkg/certificate"
 	"github.com/Venafi/vcert/v5/pkg/playbook/app/domain"
+	"github.com/Venafi/vcert/v5/pkg/playbook/app/vcertutil"
 	"github.com/Venafi/vcert/v5/pkg/playbook/util"
 	"github.com/Venafi/vcert/v5/pkg/playbook/util/capistore"
 )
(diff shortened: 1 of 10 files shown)


EduardoVV (approved):

pkg/playbook/app/installer/capi.go
```diff
@@ -47,12 +48,20 @@ func NewCAPIInstaller(inst domain.Installation) CAPIInstaller {
...
 	zap.L().Info("checking certificate health", zap.String("format", r.Type.String()), zap.String("location", r.Location))
 
-	friendlyName := request.Subject.CommonName
-	if request.FriendlyName != "" {
-		friendlyName = request.FriendlyName
+	// Get friendly name. If no friendly name is set, get CN from request as friendly name.
+	//  NOTE: This functionality is deprecated, and in a future version will be removed, and CAPIFriendlyName will be req'd
+	friendlyName := r.CAPIFriendlyName
+	if friendlyName == "" {
+		friendlyName = request.Subject.CommonName
```
luispresuelVenafi, line 55: @rvelaVenafi Shouldn't we keep the validation for CommonName being empty, which sets the FriendlyName?

pkg/playbook/app/vcertutil/vcertutil.go
```diff
@@ -241,3 +242,18 @@ func RefreshTPPTokens(config domain.Config) (string, string, error) {
 
 	return "", "", fmt.Errorf("no refresh token or certificate available to refresh access token")
 }
+
+func GeneratePassword() string {
```
luispresuelVenafi, line 246: +1

Just adding the note that in `cmd/` code we have this:
`flags.keyPassword = fmt.Sprintf("t%d-%s.tem.pwd", time.Now().Unix(), randRunes(4))`
notice that it has `tem.pwd` instead of `temp.pwd`. Ideally we want this enhanced on the other places if applicable.

luispresuelVenafi (approved):
Changes looks good. Just one minor comment. But most of a doubt.

pkg/playbook/app/vcertutil/vcertutil.go
rvelaVenafi: This is a random password being generated, that users will never see or know about. The content of the string is irrelevant

pkg/playbook/app/installer/capi.go
rvelaVenafi: Intention was never to validate the CN of the request, rather to get a FriendlyName for the CAPI stored-certificate.
It has been agreed that the value would be mapped to a property of its own: capiFriendlyName.
In order to not keep breaking compatibility, if the attribute is not set, then we take the CN from the request (old logic). A warning has been added at validation-time to make users aware they should explicitly set a FriendlyName

pkg/playbook/app/vcertutil/vcertutil.go
luispresuelVenafi: I didn't meant to let the user know about the password, rather I was suggesting we re-use this function in other places so it's more readable. Not for this PR scope though.

PR #404, opened 2023-10-07
Change certificate type to auto. Fixes #299.

### LINKED ISSUE #299: Enable support in certrequest to set the certificate type for TPP
**BUSINESS PROBLEM**
API requests to TPP's `POST certificates/request` endpoint that do not specify a certificate type default to a server certificates type in TPP.  The certificate type in TPP can have licensing implications, and so it's important to get them categorized properly. Currently, vCert SDK does not provide a way to set the certificate type, but the API endpoint in TPP allows for this.

**PROPOSED SOLUTION**
Add `CertificateType` to the certificate request struct, so that it can be optionally set by vCert, and other dependent solutions.

**CURRENT ALTERNATIVES**
None.

**VENAFI EXPERIENCE**
This beard had no gray when I started :)

BeardedPrincess:
vCert already had the proper setting to make CertificateType = "AUTO" in the TPP request. However, this was being ignored by TPP. Testing confirmed that this must be passed in all lowercase in order to work. (Probably need to file a bug with TPP for this). Changed to "auto" and tested to confirm desired behavior.  Fixes #299 .

### DIFF
--- a/pkg/venafi/tpp/connector.go
+++ b/pkg/venafi/tpp/connector.go
@@ -522,7 +522,7 @@ func prepareRequest(req *certificate.Request, zone string) (tppReq certificateRe
 		return tppReq, fmt.Errorf("Unexpected option in PrivateKeyOrigin")
 	}
 
-	tppReq.CertificateType = "AUTO"
+	tppReq.CertificateType = "auto"
 	tppReq.PolicyDN = getPolicyDN(zone)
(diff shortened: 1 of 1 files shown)


BeardedPrincess:
Additional testing proved the case of "AUTO" really doesn't make any difference. Will re-evaluate #299 with the team to understand what the issue really is.

PR #405, opened 2023-10-09
Add ability to specify a policy folder on associated device locations

### LINKED ISSUE #398: playbook: add option to create instance in specific policy folder (zone)
**BUSINESS PROBLEM**

Two scenarios that are causing a problem with our enviornment:

1. playbook runs will automatically create a device in the same policy folder as certificate. This causes an `ObjectAlreadyExists` error when the certificate name is the same as the device name. 
2. our Devices are kept in a separate policy folder than where the certificates are kept.

**PROPOSED SOLUTION**
add a `request.location.zone` option where an absolute or relative path can be added.  Relative path would have `request.zone` as the root.

**CURRENT ALTERNATIVES**
scenario 1: we can use the `request.nickname` to differentiate the object names.  This is not ideal as it takes us away from making the cert name the same as the common name.
scenario 2: we haven't found an alternative for this

BeardedPrincess:
Closes #398

When creating an associated device/application as part of a vcert playbook, you can now specify an additional `zone` that will allow for the device to be created in a separate policy folder. This is to align with best-practices which separate the policy folders for devices and certificates.


vertigo-one:
I have tested this new functionality and have been able to validate that the use of the new `location.zone` parameter works as expected. I am able to add devices and basic apps to a policy folder of my choice in the playbook independent of where the certificate object is requested.

rvelaVenafi (approved):

sbanic-venafi (approved):

PR #409, opened 2023-11-02
Updated playbook examples with associated device location

jyppy:


jyppy:
Updated playbook examples with the parameters needed to create a device object in TPP

PR #412, opened 2023-12-12
Use PBES2 with PBKDF2 and AES-256-CBC for PKCS#12 keystores

tr1ck3r:
Update password-based encryption algorithms for PKCS#12 keystores to modern level.  Should be compatible with OpenSSL 1.1.1 and higher, Java 12 and higher, and Windows Server 2019 and higher.  (See https://github.com/SSLMate/go-pkcs12/blob/master/pkcs12.go#L139-L165)


tr1ck3r:
I've added a new `--format legacy-pkcs12` option that retains the ability to generate PKCS#12 keystores using the weak algorithms.  This is consistent with the approach that was previously taken with PEM (i.e., `--format pem` for current/strong; `--format legacy-pem` for legacy/weak).

luispresuelVenafi (approved):

LaloHao (approved):

EduardoVV (approved):

PR #221, opened 2022-05-16
Fix Policy contacts multiple matching in TPP

marcos-albornoz:

PR #418, opened 2024-01-19
Contact Emails: RequestCertificate can now be given contact emails

maelvls:
To be considered:

- I wrote an end-to-end test but it will fail on VCert's TPP test instance https://supertreat.venqa.venafi.com/ because this instance only has local users, but this feature requires TPP to be configured with LDAP or AD. I'm not sure how to proceed with this.
- The func `RequestCertificate` now requires the scope `configuration` when ~`ContactEmails`~ `Contacts` is used. I don't think that's a big deal as long as we indicate it in the Go struct's documentation as well as above the RequestCertificate func.

Regarding the new ~`ContactEmails`~ `Contacts` field that I propose on the Request struct, I thought about making it more generic to allow for using an ID in the future (corresponding to the prefixed universal in TPP or a UUID in TLSPC) but since I could not find a reason for anyone to use a prefixed universal or TLSPC UUID (imagining we do implement ~`ContactEmails`~ `Contacts` for TLSPC in the future). I think it's fine to go with just ~`ContactEmails`~ `Contacts`.

Links (private to Venafi):

- [Use-case on Confluence](https://venafi.atlassian.net/wiki/spaces/~862991627/pages/2266071146/VCert+Set+the+Contact+Field+Using+an+Email+When+Requesting+Certificates+TPP)
- [Jira Feature](https://venafi.atlassian.net/browse/VC-28331)

### DIFF
--- a/pkg/certificate/request.go
+++ b/pkg/certificate/request.go
@@ -67,10 +67,10 @@ type Request struct {
 	// When an email is used by multiple TPP identities, the first identity
 	// found is picked arbitrarily.
 	//
-	// The scope `configuration` is required. Since Contact works by searching
+	// The scope `configuration` is required. Since Contacts works by searching
 	// the emails in the same LDAP or AD as the user attached to the token, you
 	// must check that you are using a user in that same identity provider.
-	// Contact doesn't work with the local TPP identities. Using Contact
+	// Contacts doesn't work with the local TPP identities. Using Contacts
 	// requires adding `mail` to the list of fields searched when performing a
 	// user search, which can be configured in the Venafi Configuration Console
 	// by RDP'ing into the TPP VM. This configuration cannot be performed
--- a/pkg/certificate/request.go
+++ b/pkg/certificate/request.go
@@ -56,8 +56,18 @@ type Request struct {
 	IssuerHint       util.IssuerHint
 
 	// ContactEmails is TPP-specific. It allows you to configure an email
-	// address to send notifications about the certificate to. Requires the
-	// scope `configuration`.
+	// address to send notifications about the certificate to. When an email is
+	// used by multiple TPP identities, the first identity found is picked
+	// arbitrarily.
+	//
+	// The scope `configuration` is required. Since ContactEmails works by
+	// searching the emails in the same LDAP or AD as the user attached to the
+	// token, you must check that you are using a user in that same identity
+	// provider. ContactEmails doesn't work with the local TPP identities. Using
+	// ContactEmails requires adding `mail` to the list of fields searched when
+	// performing a user search, which can be configured in the Venafi
+	// Configuration Console by RDP'ing into the TPP VM. This configuration
+	// cannot be performed directly in the TPP UI.
 	ContactEmails []string
 
 	// Deprecated: use ValidityDuration instead, this field is ignored if ValidityDuration is set
--- a/pkg/certificate/request.go
+++ b/pkg/certificate/request.go
@@ -55,20 +55,20 @@ type Request struct {
 	ValidityPeriod   string //represents the validity of the certificate expressed as an ISO 8601 duration
 	IssuerHint       util.IssuerHint
 
-	// ContactEmails is TPP-specific. It allows you to configure an email
+	// Contact is TPP-specific. It allows you to configure an email
 	// address to send notifications about the certificate to. When an email is
 	// used by multiple TPP identities, the first identity found is picked
 	// arbitrarily.
 	//
-	// The scope `configuration` is required. Since ContactEmails works by
+	// The scope `configuration` is required. Since Contact works by
 	// searching the emails in the same LDAP or AD as the user attached to the
 	// token, you must check that you are using a user in that same identity
-	// provider. ContactEmails doesn't work with the local TPP identities. Using
-	// ContactEmails requires adding `mail` to the list of fields searched when
+	// provider. Contact doesn't work with the local TPP identities. Using
+	// Contact requires adding `mail` to the list of fields searched when
 	// performing a user search, which can be configured in the Venafi
 	// Configuration Console by RDP'ing into the TPP VM. This configuration
 	// cannot be performed directly in the TPP UI.
-	ContactEmails []string
+	Contacts []string
 
 	// Deprecated: use ValidityDuration instead, this field is ignored if ValidityDuration is set
 	ValidityHours int
--- a/pkg/certificate/request.go
+++ b/pkg/certificate/request.go
@@ -55,19 +55,26 @@ type Request struct {
 	ValidityPeriod   string //represents the validity of the certificate expressed as an ISO 8601 duration
 	IssuerHint       util.IssuerHint
 
-	// Contact is TPP-specific. It allows you to configure an email
-	// address to send notifications about the certificate to. When an email is
-	// used by multiple TPP identities, the first identity found is picked
-	// arbitrarily.
+	// Contacts allows you to configure email addresses to send notifications
+	// about the certificate. This field is TPP-specific.
 	//
-	// The scope `configuration` is required. Since Contact works by
-	// searching the emails in the same LDAP or AD as the user attached to the
-	// token, you must check that you are using a user in that same identity
-	// provider. Contact doesn't work with the local TPP identities. Using
-	// Contact requires adding `mail` to the list of fields searched when
-	// performing a user search, which can be configured in the Venafi
-	// Configuration Console by RDP'ing into the TPP VM. This configuration
-	// cannot be performed directly in the TPP UI.
+	// Note: the user who receives the notification isn't automatically given
+	// access to that certificate. Access is configured at the policy folder
+	// level; if the user doesn't permissions on that folder, they will not be
+	// able to see the certificate's status in TPP or remediate the problem
+	// through the TPP UI.
+	//
+	// When an email is used by multiple TPP identities, the first identity
+	// found is picked arbitrarily.
+	//
+	// The scope `configuration` is required. Since Contact works by searching
+	// the emails in the same LDAP or AD as the user attached to the token, you
+	// must check that you are using a user in that same identity provider.
+	// Contact doesn't work with the local TPP identities. Using Contact
+	// requires adding `mail` to the list of fields searched when performing a
+	// user search, which can be configured in the Venafi Configuration Console
+	// by RDP'ing into the TPP VM. This configuration cannot be performed
+	// directly in the TPP UI.
 	Contacts []string
 
 	// Deprecated: use ValidityDuration instead, this field is ignored if ValidityDuration is set
--- a/pkg/certificate/request.go
+++ b/pkg/certificate/request.go
@@ -55,6 +55,11 @@ type Request struct {
 	ValidityPeriod   string //represents the validity of the certificate expressed as an ISO 8601 duration
 	IssuerHint       util.IssuerHint
 
+	// ContactEmails is TPP-specific. It allows you to configure an email
+	// address to send notifications about the certificate to. Requires the
+	// scope `configuration`.
+	ContactEmails []string
+
 	// Deprecated: use ValidityDuration instead, this field is ignored if ValidityDuration is set
 	ValidityHours int
 }
--- a/pkg/venafi/tpp/connector.go
+++ b/pkg/venafi/tpp/connector.go
@@ -1192,7 +1192,7 @@ func (c *Connector) SetPolicy(name string, ps *policy.PolicySpecification) (stri
 func (c *Connector) setContact(tppPolicy *policy.TppPolicy) (status string, err error) {
 
 	if tppPolicy.Contact != nil {
-		contacts, err := c.resolvePrefixedUniversals(tppPolicy.Contact)
+		contacts, err := c.resolveContacts(tppPolicy.Contact)
 		if err != nil {
 			return "", fmt.Errorf("an error happened trying to resolve the contacts: %w", err)
 		}
@@ -1209,28 +1209,15 @@ func (c *Connector) setContact(tppPolicy *policy.TppPolicy) (status string, err 
 	return status, nil
 }
 
-func (c *Connector) resolvePrefixedUniversals(filters []string) ([]string, error) {
-	var prefixedUniversals []string
-	identities, err := c.resolveIdentities(filters)
-	if err != nil {
-		return nil, err
-	}
-	for _, identityEntry := range identities {
-		prefixedUniversals = append(prefixedUniversals, identityEntry.PrefixedUniversal)
-	}
-
-	return prefixedUniversals, nil
-}
-
-func (c *Connector) resolveIdentities(filters []string) ([]*IdentityEntry, error) {
-	var identities []*IdentityEntry
-	uniqueContacts := getUniqueStringSlice(filters)
+func (c *Connector) resolveContacts(contacts []string) ([]string, error) {
+	var identities []string
+	uniqueContacts := getUniqueStringSlice(contacts)
 	for _, contact := range uniqueContacts {
-		identityEntry, err := c.getIdentity(contact)
+		identity, err := c.getIdentity(contact)
 		if err != nil {
 			return nil, err
 		}
-		identities = append(identities, identityEntry)
+		identities = append(identities, identity.PrefixedUniversal)
 	}
 
 	return identities, nil
--- a/pkg/venafi/tpp/connector.go
+++ b/pkg/venafi/tpp/connector.go
@@ -582,7 +582,7 @@ func (c *Connector) prepareRequest(req *certificate.Request, zone string) (tppRe
 		}
 	}
 
-	//resolving Contacts if them were provide
+	// Resolve emails to TPP identities if needed.
 	var contacts []IdentityEntry
 	if req.Contacts != nil {
 		var err error
--- a/pkg/venafi/tpp/connector.go
+++ b/pkg/venafi/tpp/connector.go
@@ -509,7 +509,7 @@ func (c *Connector) setCertificateMetadata(metadataRequest metadataSetRequest) (
 	return result.Locked, nil
 }
 
-func prepareRequest(req *certificate.Request, zone string) (tppReq certificateRequest, err error) {
+func (c *Connector) prepareRequest(req *certificate.Request, zone string) (tppReq certificateRequest, err error) {
 	switch req.CsrOrigin {
 	case certificate.LocalGeneratedCSR, certificate.UserProvidedCSR:
 		tppReq.PKCS10 = string(req.GetCSR())
@@ -581,6 +581,20 @@ func prepareRequest(req *certificate.Request, zone string) (tppReq certificateRe
 			})
 		}
 	}
+
+	//resolving Contacts if them were provide
+	var contacts []IdentityEntry
+	if req.Contacts != nil {
+		var err error
+		prefixedUniversals, err := c.resolvePrefixedUniversals(req.Contacts)
+		if err != nil {
+			return tppReq, fmt.Errorf("failed to find contact identities: %w", err)
+		}
+		for _, prefixedUniversal := range prefixedUniversals {
+			contacts = append(contacts, IdentityEntry{PrefixedUniversal: prefixedUniversal})
+		}
+	}
+	tppReq.Contacts = contacts
 
 	for name, value := range customFieldsMap {
 		tppReq.CustomFields = append(tppReq.CustomFields, customField{name, value})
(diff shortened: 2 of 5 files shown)


pkg/venafi/tpp/connector.go
```diff
@@ -1253,6 +1286,8 @@ func (c *Connector) getIdentityMatching(identities []policy.IdentityEntry, ident
 	} else {
 		return nil, fmt.Errorf("it was not possible to find the user %s", identityName)
 	}
+
+	return &(identities[0]), nil
```
marcos-albornoz, line 1290: Given the `if condition `above of this line, this will never be executed.

maelvls: Oops, I forgot to undo this change before committing. I'll undo it now.

pkg/venafi/tpp/tpp.go
```diff
@@ -63,28 +63,28 @@ type device struct {
...
-	CADN                    string          `json:",omitempty"`
-	ObjectName              string          `json:",omitempty"`
-	Subject                 string          `json:",omitempty"`
-	OrganizationalUnit      string          `json:",omitempty"`
-	Organization            string          `json:",omitempty"`
-	City                    string          `json:",omitempty"`
-	State                   string          `json:",omitempty"`
-	Country                 string          `json:",omitempty"`
-	SubjectAltNames         []sanItem       `json:",omitempty"`
-	Contact                 string          `json:",omitempty"`
```
maelvls, line 76: For some reason, the `Contact` field here was misspelled (should be `Contacts`) and it type was incorrect: it should be of the form `[{"PrefixedUniversal": "..."}]`.

pkg/venafi/tpp/tpp.go
```diff
@@ -73,7 +73,7 @@ type certificateRequest struct {
 	State                   string          `json:",omitempty"`
 	Country                 string          `json:",omitempty"`
 	SubjectAltNames         []sanItem       `json:",omitempty"`
-	Contact                 string          `json:",omitempty"`
+	Contacts                []identityReq   `json:",omitempty"`
```
maelvls, line 76: For some reason, the `Contact` field here was misspelled (should be `Contacts`) and it type was incorrect: it should be of the form `[{"PrefixedUniversal": "..."}]`.

pkg/venafi/tpp/tpp.go
```diff
@@ -73,7 +73,7 @@ type certificateRequest struct {
 	State                   string          `json:",omitempty"`
 	Country                 string          `json:",omitempty"`
 	SubjectAltNames         []sanItem       `json:",omitempty"`
-	Contact                 string          `json:",omitempty"`
+	Contacts                []contactReq    `json:",omitempty"`
```
maelvls, line 76: For some reason, the `Contact` field here was misspelled (should be `Contacts`) and it type was incorrect: it should have been of the form `[{"PrefixedUniversal": "..."}]`. 

That's why I made this change.

Context: https://docs.venafi.com/Docs/currentSDK/TopNav/Content/SDK/WebSDK/r-SDK-POST-Certificates-request.php

marcos-albornoz:
Hi Mael, I'm sharing [here a patch](https://github.com/Venafi/vcert/files/14046052/contacts.patch) with changes that are based in the changes here, the changes includes:

- The `tpp.Connector.resolveContacts()` method was refactored so it can be user not only for SetPolicy purpose but also for Request Certificate.
- The structs related to **_Identity_** were moved from _policy_ package to _tpp_ package.
- The `tpp.prepareRequest()` function was converted in a `tpp.Connector` method and it also was added the management to resolve and set the _Contacts_ to the TPP Certificate Request.

Please review the changes and if you are ok with them you can include them into your changes. I have not tested them yet but them should work.

To apply the patch to your local branch only need to checkout this branch and then use the `git apply` command in the root folder of your project as follow:
`git apply contacts.patch`

pkg/certificate/request.go
```diff
@@ -55,6 +55,21 @@ type Request struct {
...
+	//
+	// The scope `configuration` is required. Since Contact works by
+	// searching the emails in the same LDAP or AD as the user attached to the
+	// token, you must check that you are using a user in that same identity
+	// provider. Contact doesn't work with the local TPP identities. Using
+	// Contact requires adding `mail` to the list of fields searched when
+	// performing a user search, which can be configured in the Venafi
+	// Configuration Console by RDP'ing into the TPP VM. This configuration
+	// cannot be performed directly in the TPP UI.
+	Contacts []string
```
maelvls, line 71: @marcos-albornoz I am fine with renaming ContactEmails to Contacts.

Just for context, the reason I called it `ContactEmails` is because TPP's `/certificates/request` also uses the field `Contacts` but it expects something very different to emails:

```json
[
    {
        "PrefixedUniversal": "AD+prod:..."
    }
]
```

Let's go with `Contacts` then.

pkg/venafi/tpp/connector.go
```diff
@@ -1189,15 +1206,28 @@ func (c *Connector) setContact(tppPolicy *policy.TppPolicy) (status string, err
...
+
+	return prefixedUniversals, nil
+}
+
+func (c *Connector) resolveIdentities(filters []string) ([]*IdentityEntry, error) {
+	var identities []*IdentityEntry
+	uniqueContacts := getUniqueStringSlice(filters)
 	for _, contact := range uniqueContacts {
-		identity, err := c.getIdentity(contact)
+		identityEntry, err := c.getIdentity(contact)
```
maelvls, line 1226: This can't work for emails since `getIdentities` checks that the queried filter is equal to the username. When the queried filter is an email, this won't work.

maelvls:
@marcos-albornoz I turned your patch into a commit and pushed it so that we can discuss it in the PR.

About `resolveContacts` being refactored so that it can be used for emails, there is still a problem with `getIdentities`: `getIdentities` is unable to search for emails due to the check on the username that is performed by `getIdentityMatching`.

maelvls:
@marcos-albornoz I had to undo some of the changes due to `getIdentity` only working for usernames.

If only we could know why `getIdentityMatching` is checking that the queried filter matches the identity's name, we could then combine `resolveContacts` and `resolveCertificateRequestContacts`... but until then, the only possible way is to keep these funcs separate.

I have also added an end-to-end test. I had to add `t.Skip` to it because https://supertreat.venqa.venafi.com/ doesn't have an LDAP or AD identity connector configured.

marcos-albornoz:
Hi Mael, please review [here](https://github.com/Venafi/vcert/files/14061019/contacts.patch) a little update of the previous patch which I though is solving the issue in the `getIdentity()` and `getIdentityMatching()` methods.

pkg/venafi/tpp/connector.go
marcos-albornoz: please check the updated patch in my last comment, I think is solving this issue :-)

maelvls:
@marcos-albornoz I wrote unit tests for `getIdentity` in [#420](https://github.com/Venafi/vcert/pull/420). Can you review it? My goal is to add tests so that we feel feel more confident about changing `getIdentity`.

maelvls:
Thanks for the patch! I applied it to the PR (c7040c30008c4efea1383c6b90f07367aac165a9).

I propose to simplify the logic in `getIdentityMatching` by spelling out and explaining each case:

```go
switch {
case len(resp.Identities) == 0:
	return nil, fmt.Errorf("no identity found for '%s'", filter)
case len(resp.Identities) == 1:
	return &resp.Identities[0], nil
case len(resp.Identities) > 1 && !isAnEmail:
	// Since the filter isn't an email, the filter must be a username. Thus,
	// we need to ignore the results that are prefixes of the queried
	// username.
	for _, identity := range resp.Identities {
		if identity.Name == filter {
			return &identity, nil
		}
	}
	return nil, fmt.Errorf("unexpected: browseIdentities(%s) returned two identities but none of them match the username exactly", filter)
case len(resp.Identities) > 1 && isAnEmail:
	// Since it is an email, we do not need to filter out anything. So let's
	// arbitrarily return the first identity.
	return &resp.Identities[0], nil
}
```

pkg/venafi/tpp/connector.go
```diff
@@ -1231,31 +1265,53 @@ func (c *Connector) getIdentity(userName string) (*policy.IdentityEntry, error)
...
+	// Venafi Configuration Console is based on exact match, unlike `anr` used
+	// for searching usernames. Thus, we do not need to check for an exact match
+	// when an email is provided.
 
-func (c *Connector) getIdentityMatching(identities []policy.IdentityEntry, identityName string) (*policy.IdentityEntry, error) {
-	var identityEntryMatching *policy.IdentityEntry
+	// When the filter isn't an email, we consider that it is a username.
+	_, err = mail.ParseAddress(filter)
+	isEmail := err == nil
+	isUsername := !isEmail
```
marcos-albornoz, line 1288: It's not required to have two flags, only with one is enough given the origin of these values is the result of `mail.ParseAdress()` function so the flag that should keep is `isEmail` and use the resulting value to do the respective logic if the filter was an email or not

pkg/venafi/tpp/connector.go
```diff
@@ -1231,31 +1265,53 @@ func (c *Connector) getIdentity(userName string) (*policy.IdentityEntry, error)
...
 			}
 		}
-	}
-
-	//if the identity is not null
-	if identityEntryMatching != nil {
-		return identityEntryMatching, nil
-	} else {
-		return nil, fmt.Errorf("it was not possible to find the user %s", identityName)
+		return nil, fmt.Errorf("unexpected: browseIdentities(%q) returned %d identities but none of them match the username exactly", filter, len(resp.Identities))
```
marcos-albornoz, line 1304: I'm not sure if we should update the error message because we don't know if there are VCert users expecting the current error message, so for them this can cause an compatibility issue.

pkg/venafi/tpp/connector.go
```diff
@@ -1231,31 +1265,53 @@ func (c *Connector) getIdentity(userName string) (*policy.IdentityEntry, error)
...
+	case len(resp.Identities) > 1 && isEmail:
+		// The email case: we do not need to filter out anything. So let's
+		// arbitrarily return the first identity.
+		return &resp.Identities[0], nil
+	default:
```
marcos-albornoz, line 1309: I know it's a good practice to have an `default case`, but in this situation all the possible cases were covered in the previous `case` declaration so I think this `default case` should be removed because it will never executed.

pkg/venafi/tpp/connector.go
luispresuelVenafi: I'm not opposed to change the error message in favor of better understanding of the error for better UX.
I'd change this part:
`...identities but none of them match the username exactly`
to
`...identities but none of them match the username %s`
where %s is the username that was trying to be filtered. 

I'd say that filtering by username, should be assumed that value have to be exact.

Another way to highlight that value should be exact would be adding quotes:
`...identities but none of them match the username "%s"`

pkg/venafi/tpp/connector.go
maelvls: I agree, I'll revert back to just `isEmail`.

pkg/venafi/tpp/connector.go
maelvls: I agree, this `default` case is somewhat ugly. But if I remove this `default` case, I will still need to add a return statement below the switch statement; otherwise, the compiler will complain with the message "missing return statement" 😅

pkg/venafi/tpp/tpp.go
```diff
@@ -73,7 +72,7 @@ type certificateRequest struct {
 	State                   string          `json:",omitempty"`
 	Country                 string          `json:",omitempty"`
 	SubjectAltNames         []sanItem       `json:",omitempty"`
-	Contact                 string          `json:",omitempty"`
+	Contacts                []IdentityEntry `json:",omitempty"`
```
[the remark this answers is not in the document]
maelvls: Thanks, you are right, the OpenAPI spec says that the object expects in the Contacts field is the same as the IdentityEntry struct:

Screenshot 2024-01-26 at 20 19 58

pkg/venafi/tpp/connector.go
luispresuelVenafi: I liked the new proposal for the error message :+1:

pkg/venafi/tpp/connector_test.go
```diff
@@ -546,6 +548,71 @@ func TestRequestCertificateWithValidityDuration(t *testing.T) {
 	DoRequestCertificateWithValidityDuration(t, tpp)
 }
 
+func TestRequestCertificateWithContactEmails(t *testing.T) {
+	t.Skipf("Skipping test because the TPP instance doesn't have LDAP configured")
```
maelvls, line 552: I had to skip this test since https://supertreat.venqa.venafi.com/ only has local identities.

Do you think we could turn on the AD DS service on the VM that runs  https://supertreat.venqa.venafi.com/? @marcos-albornoz @luispresuelVenafi

pkg/venafi/tpp/connector.go
maelvls: Thanks for the patch. I now understand better why the test `identity.Name == filter` has been introduced a while back. I detailed this in https://github.com/Venafi/vcert/pull/221#issuecomment-1911747389.

maelvls:
@luispresuelVenafi @marcos-albornoz Thank you for your reviews! I addressed all of your comments, please take another look

pkg/venafi/tpp/connector.go
maelvls: I removed the `default` statement, and moved the return statement to the end of the function.

pkg/venafi/tpp/connector.go
```diff
@@ -1231,31 +1265,51 @@ func (c *Connector) getIdentity(userName string) (*policy.IdentityEntry, error)
...
+		return &resp.Identities[0], nil
 	}
 
-	//if the identity is not null
-	if identityEntryMatching != nil {
-		return identityEntryMatching, nil
-	} else {
-		return nil, fmt.Errorf("it was not possible to find the user %s", identityName)
-	}
+	return nil, fmt.Errorf("programmer mistake: impossible case in getIdentity")
```
marcos-albornoz, line 1309: Please change this line for the following or something similar 
`return nil, fmt.Errorf("it was not possible to find an Identity using filter value %s", filter)`

maelvls: I suggest prefixing this error with "programmer mistake" because this error isn't meant to happen. If this error occurs, it means that I made a mistake when programming this. "programmer mistake" tells the user that it isn't their fault.

The message "it was not possible to find an Identity using this filter" doesn't describe what has happened. Here, the filter may have returned identities, but none of the switch cases have matched, which is unexpected since these 4 switch cases are meant to catch 100% of the cases. 

I propose the following error; it should help us (developers) debug the issue if it ever happen:

```
programmer mistake: browseIdentities returned 2 results for filter 'jsmith' and none of the switch cases matched
```

Note that panicking would be a good fit here, but panicking isn't a great experience for the callers of VCert, so let's not panic.

maelvls:
**⚠️ NOTE ⚠️:** Please suspend reviewing this PR until I've had the chance to write some unit tests around `getIdentity` using the old implementation from the master branch. I am writing these tests in https://github.com/Venafi/vcert/pull/420. Once #420 is merged into master, we can resume reviewing #418. You'll notice that none of the changes from #418 are included in #420, which should prevent any potential regressions.

maelvls:
Marcos caught a regression that I had introduced with my `switch` statement: when searching for `foo`, if the API simply returns `foobar`, the code would mistakenly return the identity `foobar` instead of returning an error.

I have added a test case to #420 and fixed the regression in 2340938.

pkg/venafi/tpp/connector_test.go
marcos-albornoz: Let's leave this off, only modify the message to say something like "skipping by default because this test only works over TPP instances which have LDAP users configured."

marcos-albornoz (approved):

pkg/venafi/tpp/connector_test.go
maelvls: Thanks, I have updated the message to say that.

maelvls:
The build passed.

https://jenkins.eng.venafi.com/job/VCert/job/vcert/3178/

Screenshot 2024-01-30 at 19 39 16

rvelaVenafi (approved):
Code looks good.

pkg/venafi/tpp/connector.go
```diff
@@ -1231,31 +1265,51 @@ func (c *Connector) getIdentity(userName string) (*policy.IdentityEntry, error)
...
 
-	//if the identity is not null
-	if identityEntryMatching != nil {
-		return identityEntryMatching, nil
-	} else {
-		return nil, fmt.Errorf("it was not possible to find the user %s", identityName)
-	}
+	// The above switch cases must catch 100% of the cases. If we arrive here,
+	// it means that we have made a programming mistake.
+	return nil, fmt.Errorf("programmer mistake: browseIdentities returned %d identities for the filter '%s' and none of the switch cases matched, but the switch cases are expected to catch 100%% of the cases", len(resp.Identities), filter)
```
maelvls, line 1309:
```suggestion
	return nil, fmt.Errorf("this was not supposed to happen, please report to the developer team: browseIdentities returned %d identities for the filter '%s' and none of the switch cases matched, but the switch cases are expected to catch 100%% of the cases", len(resp.Identities), filter)
```

pkg/venafi/tpp/connector_test.go
```diff
@@ -3193,15 +3258,15 @@ func Test_getIdentity(t *testing.T) {
 				Type:              1,
 				Universal:         "{c1f0ac9a-c9da-55a7-963a-fd6445af8307}",
 			}},
-			wantErr: "it was not possible to find the user foo",
+			wantErr: "1 identities were found but none of these identities has a username equal to 'foo'",
```
maelvls, line 3261: This is a breaking change.

maelvls:
Good news, `setpolicy` now works with emails! Previously, it was only possible to use a username (e.g., `mael.valais`).

For example, let's imagine I want to configure the Contacts field of a given policy folder, I can do `vcert setpolicy` with an email:

```yaml
{
  "users": [
    "mael.valais@venafi.com"
  ],
  "policy": {
    "wildcardAllowed": true,
    "certificateAuthority": "\\VED\\Policy\\Administration\\CAs\\Microsoft CA Web Server 1 Year",
    "keyPair": {
      "reuseAllowed": false
    },
    "subjectAltNames": {
      "dnsAllowed": true,
      "ipAllowed": true,
      "emailAllowed": true,
      "uriAllowed": true,
      "upnAllowed": true
    }
  },
  "defaults": {
    "keyPair": {
      "keyType": "RSA",
      "rsaKeySize": 2048,
      "serviceGenerated": false
    },
    "autoInstalled": false
  }
}
```

Unfortunately, `vcert getpolicy` will still return a username:

```yaml
{
  "users": [
    "mvalais"
  ],
  "policy": {
    "wildcardAllowed": true,
    "certificateAuthority": "\\VED\\Policy\\Administration\\CAs\\Microsoft CA Web Server 1 Year",
    "keyPair": {
      "reuseAllowed": false
    },
    "subjectAltNames": {
      "dnsAllowed": true,
      "ipAllowed": true,
      "emailAllowed": true,
      "uriAllowed": true,
      "upnAllowed": true
    }
  },
  "defaults": {
    "keyPair": {
      "keyType": "RSA",
      "rsaKeySize": 2048,
      "serviceGenerated": false
    },
    "autoInstalled": false
  }
}
```

maelvls:
I fixed the two problems we talked about in the meeting. This PR is now ready to be merged!

luispresuelVenafi:
Just validated the new commits. Looks Good Go Me

pkg/certificate/request.go
```diff
@@ -55,6 +55,28 @@ type Request struct {
...
+	// found is picked arbitrarily.
+	//
+	// The scope `configuration` is required. Since Contacts works by searching
+	// the emails in the same LDAP or AD as the user attached to the token, you
+	// must check that you are using a user in that same identity provider.
+	// Contacts doesn't work with the local TPP identities. Using Contacts
+	// requires adding `mail` to the list of fields searched when performing a
+	// user search, which can be configured in the Venafi Configuration Console
+	// by RDP'ing into the TPP VM. This configuration cannot be performed
+	// directly in the TPP UI.
```
maelvls, line 77: This documentation mistakenly talks about emails, but emails aren't what we recommend using. Customers who uses SSO will have configured their AD search with `userPrincipalName` and not `mail`. This documentation should rather say that contacts are a list of identity strings that matches one of the AD or LDAP search attributes; it should also say that most of the time, these attributes are `anr` and `userPrincipalName`.

PR #420, opened 2024-01-26
Contact Emails: Add unit tests for `getIdentity`

maelvls:
This PR should be merged before continuing reviews on #418.

I paused working on #418 so that I could write some unit tests around `getIdentity` with the existing implementation (the one in the master branch). Once we have the unit tests merged to master, we can resume working on #418.

I could have added all of these tests directly to #418, but we wouldn't be sure whether the old implementation and the new implementation pass the same tests.

This way, we can decrease the chance of introducing a regression in `getIdentity`.

To know more about the tricks that `getIdentity` does, you can read https://github.com/Venafi/vcert/pull/221#issuecomment-1911747389.

### DIFF
--- a/pkg/venafi/tpp/connector_test.go
+++ b/pkg/venafi/tpp/connector_test.go
@@ -28,6 +28,7 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
@@ -48,6 +49,7 @@ import (
 	"github.com/Venafi/vcert/v5/pkg/util"
 	"github.com/Venafi/vcert/v5/pkg/verror"
 	"github.com/Venafi/vcert/v5/test"
+	"github.com/stretchr/testify/require"
 )
 
 var ctx *test.Context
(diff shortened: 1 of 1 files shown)


maelvls:
This is ready to be reviewed.

luispresuelVenafi (approved):
LGTM

marcos-albornoz:
> @marcos-albornoz Could you take a look? I think I need two reviews to get this merged. Thanks!

I've reviewed it but it's seeming this code is out of date compared with the code in [PR 418](https://github.com/Venafi/vcert/pull/418) because this code is still referencing to the `policy.IdentityEntry` given that struct was moved to `tpp` package. Also I see in the [PR 418](https://github.com/Venafi/vcert/pull/418) there is a more extended version of the test in this PR

maelvls:
This PR only focuses on adding the missing unit tests around `getIdentity` without changing anything else in VCert. That's why `IdentityEntry` is still in the `policy` folder.

You are correct, I have added more test cases in https://github.com/Venafi/vcert/pull/418 to account for the new behavior in `getIdentity`.

Does it make sense? I am trying to add unit tests before changing everything to lower the chance of mistakenly breaking something.

maelvls:
Marcos discovered a regression I introduced in #420: when `foo` is searched and the only result is `foobar`, `getIdentity` would mistakenly return `foobar` instead of returning an error.

I have added a test case to cover this. Please take another look @marcos-albornoz

maelvls:
I'll rebase the pass the requirement "The base branch requires all commits to be signed."

maelvls:
Success!

https://jenkins.eng.venafi.com/job/VCert/job/vcert/3176/

Screenshot 2024-01-30 at 19 03 07

PR #424, opened 2024-02-13
Allow access_token to be passed to TLSPC / VaaS from playbook

BeardedPrincess:
Allow the vCert playbook to specify an access_token for TLSPC.  This does not break any existing functionality, or directly allow using an access_token. However, some use-cases have identified the requirement for the use of an API gateway between the client and VaaS, along with the ability to pass a bearer token.  This fix allows that functionality for that corner case. Additionally, it prepares us for updates coming to allow "service account" usage for vCert.

### DIFF
--- a/examples/tlspc-svc-account/main.go
+++ b/examples/tlspc-svc-account/main.go
@@ -58,19 +58,23 @@ func main() {
 		log.Fatalf("error creating client: %s", err.Error())
 	}
 
-	_, err = connector.ReadZoneConfiguration()
+	zoneConfig, err := connector.ReadZoneConfiguration()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


Pmaraveyias (approved):
Changes look good

rvelaVenafi:
@BeardedPrincess with the service account feature support, is this PR still valid?

BeardedPrincess:
Yes, it will still be needed. The use-case is still probably more of an edge case where we need to operate with an API gateway. If you want to reach out I can spend some time explaining the use-case.

The service-account support does mitigate this, but unfortunately, I'm not sure it eliminates it 100%

PR #433, opened 2024-02-28
Enables WorkToDoTimeout for certificate request

luispresuelVenafi:
In order for other integrations to consume and override enrollment timeout. We enable `WorkDoTimeout` in order to override TLSPDC (TPP) default CA timeout (60secs). To be noted that as of now, maximum value is 120 seconds. If greater value is provided (e.g. 180 secs), TPP will automatically take 120 secs.

### DIFF
--- a/pkg/certificate/request.go
+++ b/pkg/certificate/request.go
@@ -47,7 +47,16 @@ type Request struct {
 	FetchPrivateKey bool
 	/*	Thumbprint is here because *Request is used in RetrieveCertificate().
 		Code should be refactored so that RetrieveCertificate() uses some abstract search object, instead of *Request{PickupID} */
-	Thumbprint       string
+	Thumbprint string
+	// Timeout we have multiple purposes for timeout (bad practice, we need to correct this in the future)
+	// TPP (a.k.a TLPSDC): we use it in order to set WorkToDoTimeout, that overrides TPP default timeout waiting time for the CA to finish
+	// if the value is more than the maximum value, TPP will automatically set the maximum value supported (as of the moment of this
+	// commit, 120 seconds).
+	// Cloud (a.k.a VaaS a.k.a TLPSC) : We use this timeout in our RetrieveCertificate function which handles a retry logic
+	// TPP SSH feature: We override the http client default timeout to perform http requests.
+	// Firefly: not usage at all
+	// VCert CLI: We have hardcoded 180 seconds for retrieve certificate operation. For VaaS it will set retry logic for
+	// 180 seconds and TPP will override CA timeout as the hardcoded value
 	Timeout          time.Duration
 	CustomFields     []CustomField
 	Location         *Location
--- a/pkg/certificate/request.go
+++ b/pkg/certificate/request.go
@@ -48,14 +48,16 @@ type Request struct {
 	/*	Thumbprint is here because *Request is used in RetrieveCertificate().
 		Code should be refactored so that RetrieveCertificate() uses some abstract search object, instead of *Request{PickupID} */
 	Thumbprint string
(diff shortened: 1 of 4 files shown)


pkg/certificate/request.go
```diff
@@ -47,7 +47,16 @@ type Request struct {
 	FetchPrivateKey bool
 	/*	Thumbprint is here because *Request is used in RetrieveCertificate().
 		Code should be refactored so that RetrieveCertificate() uses some abstract search object, instead of *Request{PickupID} */
-	Thumbprint       string
+	Thumbprint string
+	// Timeout we have multiple purposes for timeout (bad practice, we need to correct this in the future)
```
rvelaVenafi, line 51: If this is something we should not encourage. Why adding another meaning and purpose to this attribute instead of creating a new one?

pkg/venafi/tpp/connector.go
```diff
@@ -653,6 +653,14 @@ func (c *Connector) prepareRequest(req *certificate.Request, zone string) (tppRe
 	//    - true: Clear the Disabled attribute, reenable, and then renew the certificate (in this request). Reuse the same CertificateDN, that is also known as a Certificate object.
 	tppReq.Reenable = true
 
+	// If enable timeout is defined by the user in the request, we use it in order
+	// override API's timeout for the CA to finish issuance
+	if req.Timeout != 0 {
```
rvelaVenafi, line 658: Change it to `if req.Timeout > 0` to rule out negative values

rvelaVenafi (changes requested):

pkg/venafi/tpp/connector.go
```diff
@@ -653,6 +653,14 @@ func (c *Connector) prepareRequest(req *certificate.Request, zone string) (tppRe
 	//    - true: Clear the Disabled attribute, reenable, and then renew the certificate (in this request). Reuse the same CertificateDN, that is also known as a Certificate object.
 	tppReq.Reenable = true
 
+	// If enable timeout is defined by the user in the request, we use it in order
```
rvelaVenafi, line 656: make sure comment matches the actual code logic

rvelaVenafi (changes requested):

pkg/certificate/request.go
luispresuelVenafi: Ideally we need to standardize this, as there a many different usage, but is not the purpose of this MR hence I mentioned what we already:
> bad practice, we need to correct this in the future

Without any standardization **as user**, I'd expect the request attribute `Timeout`, at very least serve, the purpose, that allows me to increase the timeout for enrolling certificate, which it is what I'm proposing we do for TLSPDC side of things. (As of now, that attribute has no use in TLPSDC for certificate enrollment).

pkg/certificate/request.go
```diff
@@ -47,7 +47,16 @@ type Request struct {
...
+	// TPP (a.k.a TLPSDC): we use it in order to set WorkToDoTimeout, that overrides TPP default timeout waiting time for the CA to finish
```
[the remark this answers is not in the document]
luispresuelVenafi: Done by latest commit

pkg/certificate/request.go
luispresuelVenafi: Done by latest commit

pkg/venafi/tpp/connector.go
luispresuelVenafi: Updated comment by latest commit

pkg/venafi/tpp/connector.go
luispresuelVenafi: Done by latest commit

sbanic-venafi (approved):

PR #435, opened 2024-03-12
feat(svc-account-auth): Adds support for service account authentication on TLSPC (VCert SDK)

rvelaVenafi:
Adds new authentication method for TLSPC connector: service account.

The Authentication struct has two new attributes:

`TenantID`: the ID of the tenant that is requesting the access token from TLSPC
`ExternalIdPJWT`: JWT obtained from the Identity Provider associated with the service account being used. It will be used to retrieve a new access token

Closes VC-31642


sbanic-venafi (approved):

pkg/venafi/cloud/connector.go
```diff
@@ -306,47 +311,46 @@ func (c *Connector) GetPolicy(name string) (*policy.PolicySpecification, error)
...
+					return nil, err
 				}
-				if teams != nil {
-					for _, team := range teams.Teams {
-						if team.ID == owner.OwnerId {
-							usersList = append(usersList, team.Name)
-							break
-						}
+			}
+			if teamsList != nil {
```
luispresuelVenafi, line 344: We don't need this validation if you already validated that `teamsList` is nil above.

pkg/endpoint/authentication.go
```diff
@@ -19,15 +19,27 @@ package endpoint
 // Authentication provides a struct for authentication data. Either specify User and Password for Trust Protection Platform
 // or Firefly or ClientId and ClientSecret for Firefly or specify an APIKey for TLS Protect Cloud.
 type Authentication struct {
-	User         string `yaml:"user,omitempty"`
-	Password     string `yaml:"password,omitempty"`
-	APIKey       string `yaml:"apiKey,omitempty"`
+	//TPP Auth methods
+	// User and password
+	User     string `yaml:"user,omitempty"`     //**DEPRECATED** Use access/refresh token or client certificate instead
+	Password string `yaml:"password,omitempty"` //**DEPRECATED** Use access/refresh token or client certificate instead
```
luispresuelVenafi, line 25: +1

PR #439, opened 2024-03-20
Fixes to support service account authentication in TLSPC

rvelaVenafi:
Closes VC-31642

### DIFF
--- a/pkg/venafi/cloud/search.go
+++ b/pkg/venafi/cloud/search.go
@@ -181,7 +181,7 @@ func formatSearchCertificateArguments(cn string, sans *certificate.Sans, certMin
 				{
 					Field:    "validityPeriodDays",
 					Operator: GTE,
-					Value:    certMinTimeDays,
+					Value:    &certMinTimeDays,
 				},
 			},
 		},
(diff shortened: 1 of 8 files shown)


pkg/venafi/cloud/search.go
```diff
@@ -181,7 +181,7 @@ func formatSearchCertificateArguments(cn string, sans *certificate.Sans, certMin
 				{
 					Field:    "validityPeriodDays",
 					Operator: GTE,
-					Value:    certMinTimeDays,
```
luispresuelVenafi, line 184: Are you sure about this change? I'm concerned whenever we update the PKI Backend plugin. I recall before we had to deal with pointer issues due to the value not being updated properly during "search valid certificate" feature

rvelaVenafi: Im sure it didnt work before, the parameter is declared as an interface, interfaces are pointers. When assigned, it passed the pointer address instead of the value

luispresuelVenafi: I'm not saying it shouldn't not be a pointer, I recall that for PKI Backend we discovered a pointer issue coming from this functionality and we had to "hack" it to make sure that value is properly updated within a for loop. We don't have a for loop in here but just bringing some concern since this is working nice as it is as of now (before this change)

rvelaVenafi: the function will not change, as the sanDNS behavior is not modified. Only makes work the attributes of the search: the cn and the minimum validity days

EduardoVV (approved):

luispresuelVenafi (approved):

PR #440, opened 2024-03-21
WIP: [VC-31275] Allow user to configure the User-Agent HTTP header in vcert CLI

wallrj:
Depends on: #443 

> After discussing with @maelvls I have spun out the SDK changes to #443.
> Because this `vcert --user-agent` feature is a nice-to-have and is not required for VC-31275.

I've added a `--user-agent` flag which allows you to override the User-Agent header that is added to requests made by the `vcert` CLI.
That can also be configured by a new environment variable `VCERT_USER_AGENT`.
`vcert` CLI will now use `User-Agent: vCert/v5.6.0` when using a released binary, where the version has been embedded
(or `vCert/Unknown` when compiled locally).

### Validation of `--user-agent`
```sh
go run ./cmd/vcert renew \
  -k foo --no-prompt -id foo --trust-bundle ~/.mitmproxy/mitmproxy-ca.pem \
  --user-agent richard.$'\x7F'.wall 
```

```sh
vCert: 2024/03/22 17:27:17 Invalid user-agent value: "richard.\x7f.wall"
exit status 1
```

### With mitmproxy
* https://mitmproxy.org/
```sh
mitmproxy
```
```sh
export HTTPS_PROXY=localhost:8080 
```

* Command line: `--user-agent`
```sh
go run ./cmd/vcert renew   \
  -k foo --no-prompt -id foo --trust-bundle ~/.mitmproxy/mitmproxy-ca.pem \
  --user-agent my-vcert-script/v1.2.3
```

* Environment Variable: `VCERT_USER_AGENT`
```sh
$ VCERT_USER_AGENT=env-user-agent/v2.3.4 go run ./cmd/vcert renew     -k foo --no-prompt -id foo --trust-bundle ~/.mitmproxy/mitmproxy-ca.pem                                                                              

```

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -20,15 +20,24 @@ import (
 	"crypto/tls"
 	"crypto/x509/pkix"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/http/httputil"
 	"os"
+	"strings"
 	"testing"
 	"time"
 
+	"github.com/go-http-utils/headers"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/utils/ptr"
+
 	"github.com/Venafi/vcert/v5/pkg/certificate"
 	"github.com/Venafi/vcert/v5/pkg/endpoint"
+	"github.com/Venafi/vcert/v5/pkg/util"
 )
 
 func init() {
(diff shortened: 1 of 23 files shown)


cmd/vcert/validators.go
```diff
@@ -75,6 +76,9 @@ func readData(commandName string) error {
 
 func validateCommonFlags(commandName string) error {
 
+	if !httpguts.ValidHeaderFieldValue(flags.userAgent) {
+		return fmt.Errorf("Invalid user-agent value: %q", flags.userAgent)
+	}
```
wallrj, line 81: This is the same validation that is performed by the Go http.Client before it writes the headers to the network:
 * https://github.com/golang/go/blob/5f5b20c4268c1a3aa6a3b132aeede6dc82adf344/src/net/http/transport.go#L516-L530

client_test.go
```diff
@@ -128,3 +134,223 @@ func TestNewClientWithFileConfig(t *testing.T) {
...
+		},
+		endpoint.ConnectorTypeTPP: endpoint.Authentication{
+			User:     "fake-user",
+			Password: "fake-password",
+		},
+		endpoint.ConnectorTypeFirefly: endpoint.Authentication{
+			IdentityProvider: &endpoint.OAuthProvider{
+				DeviceURL: "https://device.oauth.example.local",
+			},
+		},
```
wallrj, line 232: If you change this to use  `TokenURL` you find that the Firefly connector is not using the vcert.Client http Client, but is using calling an oauth2 function which uses the `http.DefaultClient`.

So the requests to the TokenURL will not have the standard User-Agent.
* https://github.com/golang/oauth2/issues/179

There is a way to set it, by adding our http.Client to the context and then adding a RoundTripper wrapper which injects the User-Agent, but I propose postponing that for another PR.

cmd/vcert/flags.go
```diff
@@ -694,7 +696,16 @@ var (
...
 
-	commonFlags              = []cli.Flag{flagInsecure, flagVerbose, flagNoPrompt}
+	flagUserAgent = &cli.StringFlag{
+		Name: "user-agent",
+		Usage: "Use to set the User-Agent header in HTTP requests to the Venafi REST API. " +
+			"Example: --user-agent example-vcert-script/v0.0.0",
+		Destination: &flags.userAgent,
+		Value:       fmt.Sprintf("%s/%s", UtilityShortName, vcert.GetFormattedVersionString()),
+		EnvVars:     []string{vcertUserAgent},
+	}
```
wallrj, line 706: The other flags don't seem to make use of the `EnvVars` field, but it's nice because it helps the user learn about the corresponding env var name in the `--help` output.

client_test.go
BeardedPrincess: Agreed. There's no analytics on Firefly at the moment that would take advantage of this yet anyway, but maybe something we can ask @tr1ck3r to determine if there's value there.  Agree to leave it to another PR after we reconcile that part.

cmd/vcert/flags.go
BeardedPrincess: Good callout, we should look to update the other commands in a future effort.

maelvls:
Hey, really glad to see that this long lasting issue will be solved soon!

Why would one have to set `--user-agent` or `VCERT_USER_AGENT`? In my mind, the CLI should have a sane default such as `User-Agent: vCert/v5.6.0` and the user shouldn't have to use some flag to enable that. Are there reasons (such as not breaking existing flows) that would support the decision of keeping the old behavior by default?

Also, nit: is the official spelling `vCert` or `VCert`? I've only seen `vcert` (when speaking about the CLI binary) and `VCert` (when speaking about the product) in this repository and in https://support.venafi.com/hc/en-us/articles/217991528-Introducing-VCert-API-Abstraction-for-DevOps.

cmd/vcert/flags.go
rvelaVenafi: There is reason behind that decision. We need to sit down and think about it. So far the only reason I remember VCert CLI does not use native EnvVar support is to be able to discern the source of the flag (cli, envvar, config file) and print an appropiate message (flag X is overriden by cli...)

rvelaVenafi:
@maelvls So far, the official documentation (readme file) has VCert as the official name. This info pre-dates me as maintainer of the repo, so thats how much I know.

cmd/vcert/flags.go
rvelaVenafi: BTW, Im all in favor to drop the custom env var support we have in favor of the native CLI support.

rvelaVenafi:
@wallrj First of all, thanks a lot for this contribution, we at the Open Source team really appreciate it. 

Now, I apologize for the inconvenience, but there are some big changes coming to VCert this week. We are adding support for service account authentication in TLSPC (VaaS) connector. These changes cover a wide range of files across the SDK, CLI and playbook packages. 

We are also updating the go version to 1.21.

So, I would recommend to wait after those changes are merged and then rebase your branch. Im quite sure a couple of things will need fixing (due to collisions, like flags and envars).

Once again, thank you for your contribution and I hope this mildly annoying turn of events do not discourage you from contributing in the future

wallrj:
> Why would one have to set `--user-agent` or `VCERT_USER_AGENT`? In my mind, the CLI should have a sane default such as `User-Agent: vCert/v5.6.0` and the user shouldn't have to use some flag to enable that. Are there reasons (such as not breaking existing flows) that would support the decision of keeping the old behavior by default?

The default User-Agent header for the `vcert` CLI is `vCert/v5.6.0`. It is defined in `flags.go`. 
You can see an example of the default header in the testing section of the PR description.

You can _override_ the default User-Agent header of the CLI using the `--user-agent` flag or with the `VCERT_USER_AGENT` environment variable.
For example, if you execute `vcert` from a shell or Python script, you might set `VCERT_USER_AGENT` to help the Venafi TPP administrator distinguish your software from other utilities that might also be wrapping the `vcert` CLI.

I will add this explanation to the README file.

The default User-Agent header for the vcert _Go module_, is still `Go-HTTP-1.1`  and that might be a mistake.
In the PR description, I argued that this is for "backwards compatibility"....but  that isn't consistent with my changing the  default for the CLI. 
Perhaps the defaults should be:
* `vcert/x.y.z` for `vcert` CLI
* `vcert-go/<git-commit-hash>` for vcert Go module.

Software using the Go module (such as cert-manager) can  change it by setting the `client.Config.UserAgent` field when instantiating a `vcert.Client`.

Software should also be able to configure the client to **omit** the User-Agent by setting the value to empty string,
and for that I will need to change the field to a pointer so that I can distinguish between empty string and no preference.

> Also, nit: is the official spelling `vCert` or `VCert`? I've only seen `vcert` (when speaking about the CLI binary) and `VCert` (when speaking about the product) in this repository and in https://support.venafi.com/hc/en-us/articles/217991528-Introducing-VCert-API-Abstraction-for-DevOps.

I agree that the spelling looks wrong, but it has been defined in a constant and used in logs and `--help` since 2018. 
It can be changed in a separate PR.
 * https://github.com/search?q=repo%3AVenafi%2Fvcert%20UtilityShortName&type=code
 * https://github.com/Venafi/vcert/commit/f2292976060c9054a85ebe05015409d830b69057#diff-972f036a958aeee05a09eecf4a9dc3d26dfbabb12187f69c158810b2089ee2a1R25
 * https://github.com/Venafi/vcert/commit/39a089993f5723a8b740a9b1a4dcf4bac0a73e65

wallrj:
After discussing with @maelvls I have spun out the SDK changes to https://github.com/Venafi/vcert/pull/443.
Because this vcert --user-agent feature is a nice-to-have and is not required for VC-31275.

Issue #437, opened 2024-03-20
Set the User-Agent HTTP header in requests to Venafi API server


wallrj:
**BUSINESS PROBLEM**

As an administrator of a Venafi TPP server or of a Venafi Cloud tenant, I want to know which software is being used to connect to the REST API.

Right now the `vcert` CLI and any software that uses the vcert SDK will use the [default Go HTTP client User-Agent header](https://github.com/golang/go/blob/db423dde85ad4923c2c4addb1cd96f119c7b6dc6/src/net/http/request.go#L534-L538): `Go-http-client/1.1` and the requests will be indistinguishable.

For example, if the TPP API server is being overwhelmed by a rapid repeated requests, I would like to be able to look at the IIS server logs and see from the User-Agent of the offending requests and that the requests *may* be being generated by a `vcert` command because I see `User-Agent: vcert/vX.Y.Z`.

As a maintainer of the cert-manager Venafi Issuer, I want to be able to use the vcert SDK and configure it to send `User-Agent: cert-manager/vX.Y.Z` so that the administrator of a Venafi TPP server or Venafi Cloud service, can know which software is being used to connect to the REST API. Why? Because if there is a bug in the cert-manager Venafi issuer, which causes it to not back off in the event of a failed API request, the Venafi server administrator can quickly know that cert-manager is the culprit and quickly report the bug so that it be quickly fixed.

**PROPOSED SOLUTION**

Add a default User-Agent HTTP header to the Vcert client.
Allow the User-Agent to be **easily** overridden by software that imports the Vcert SDK.

**CURRENT ALTERNATIVES**

To change the User-Agent of `vcert` CLI requires modification of the source code and recompilation.

To change the User-Agent when importing the vcert SDK requires supplying a custom `http.Client` with a [RoundTripper wrapped transport](https://github.com/cert-manager/cert-manager/pull/6865):
* https://github.com/Venafi/vcert/blob/a30535acfbeac7a1d5d1c8fe450a772d6c6faa68/config.go#L64-L65
* https://github.com/cert-manager/cert-manager/pull/6865

<img width="960" alt="image" src="https://github.com/Venafi/vcert/assets/978965/73cc91c8-afd2-4885-91cd-895c19f908f0">

BeardedPrincess:
@wallrj I think this a valuable change for the reasons you have listed. It would also be useful to be able to identify clients and applications that are using a potentially old version of the SDK.

Do you anticipate that you will be able to submit a PR for these?

PR #443, opened 2024-03-27
[VC-31275] Allow user to configure the User-Agent HTTP header in vcert SDK

wallrj:
**Software like cert-manager can now supply its own User-Agent when it instantiates the vcert.Client.**
> Why? Because if there is a bug in the cert-manager Venafi issuer, which causes it to not back off in the event of a failed API request, the Venafi server administrator can quickly know that cert-manager is the culprit and quickly report the bug so that it be quickly fixed.

**The default User-Agent for `vcert` CLI and for any software importing the vcert/v5 is now: `vcert/v5`.**
> Why? Because right now the vcert CLI and any software that uses the vcert SDK will use the [default `Go-http-client/1.1` User-Agent header](https://github.com/golang/go/blob/db423dde85ad4923c2c4addb1cd96f119c7b6dc6/src/net/http/request.go#L534-L538) and the requests will be indistinguishable from any other Go http client.
The new default header will allow the Venafi administrator to estimate the adoption of vcert v5.

Fixes: https://github.com/Venafi/vcert/issues/437

* https://github.com/Venafi/vcert/issues/437
* VC-31275

### `vcert` CLI default
```sh
export HTTPS_PROXY=localhost:8080
go run ./cmd/vcert renew     -k foo --no-prompt -id foo --trust-bundle ~/.mitmproxy/mitmproxy-ca.pem
```

```sh
vCert: 2024/03/27 10:54:54 Warning: --platform not set. Attempting to best-guess platform from connection flags
vCert: 2024/03/27 10:54:54 Detected trust bundle...
vCert: 2024/03/27 10:54:54 You specified a trust bundle.
vCert: 2024/03/27 10:54:55 Unable to connect to Venafi as a Service: vcert error: server error: 401 Unauthorized
vCert: 2024/03/27 10:54:55 Failed to fetch old certificate by id foo: must be autheticated to request a certificate
exit status 1
```

image

### Examples
I've partially tested the examples, in so far as I ran each example with some fake settings in environment variables and observed that the expected User-Agent header was present in the first failing request.

* firefly
image

* server
image

* simple-cli
image

* tlspc-svc-account
image

### DIFF
--- a/pkg/venafi/cloud/connector.go
+++ b/pkg/venafi/cloud/connector.go
@@ -32,6 +32,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-http-utils/headers"
 	"golang.org/x/crypto/nacl/box"
 
 	"github.com/Venafi/vcert/v5/pkg/certificate"
@@ -90,12 +91,13 @@ type Connector struct {
 	trust       *x509.CertPool
 	zone        cloudZone
 	client      *http.Client
+	userAgent   string
 }
 
 // NewConnector creates a new Venafi Cloud Connector object used to communicate with Venafi Cloud
 func NewConnector(url string, zone string, verbose bool, trust *x509.CertPool) (*Connector, error) {
 	cZone := cloudZone{zone: zone}
-	c := Connector{verbose: verbose, trust: trust, zone: cZone}
+	c := Connector{verbose: verbose, trust: trust, zone: cZone, userAgent: util.DefaultUserAgent}
 
 	var err error
 	c.baseURL, err = normalizeURL(url)
@@ -112,6 +114,10 @@ func (c *Connector) GetType() endpoint.ConnectorType {
 func (c *Connector) SetZone(z string) {
 	cZone := cloudZone{zone: z}
 	c.zone = cZone
+}
+
+func (c *Connector) SetUserAgent(userAgent string) {
+	c.userAgent = userAgent
 }
 
 func (c *Connector) SetHTTPClient(client *http.Client) {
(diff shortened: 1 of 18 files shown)


wallrj:
/cc @rvelaVenafi 

Please review.

pkg/venafi/cloud/connector.go
```diff
@@ -39,6 +39,7 @@ import (
 	"github.com/Venafi/vcert/v5/pkg/policy"
 	"github.com/Venafi/vcert/v5/pkg/util"
 	"github.com/Venafi/vcert/v5/pkg/verror"
+	"github.com/go-http-utils/headers"
```
rvelaVenafi, line 42: please move this import up. The import block format should be as follow:

```
native imports

third party imports

vcert imports
```

pkg/venafi/firefly/firefly.go
```diff
@@ -33,6 +33,7 @@ import (
 	"github.com/Venafi/vcert/v5/pkg/certificate"
 	"github.com/Venafi/vcert/v5/pkg/endpoint"
 	"github.com/Venafi/vcert/v5/pkg/verror"
+	"github.com/go-http-utils/headers"
```
rvelaVenafi, line 36: same as previous comment. Please sort imports

pkg/venafi/tpp/tpp.go
```diff
@@ -33,6 +33,7 @@ import (
 
 	"github.com/Venafi/vcert/v5/pkg/certificate"
 	"github.com/Venafi/vcert/v5/pkg/endpoint"
+	"github.com/go-http-utils/headers"
```
rvelaVenafi, line 36: same s previous comments. Please sort imports

rvelaVenafi (changes requested):

rvelaVenafi (approved):

JigarAtVenafi (approved):

PR #448, opened 2024-04-04
Adds tokenURL attribute for VCP service account authentication

rvelaVenafi:
VCert SDK:
- Removes TenantID from endpoint.Authentication struct
- Cloud connector will use endpoint.Authentication.OAuthProvider.TokenURL instead of building the URL using the tenantID to obtain the access token

VCert CLI:
- Removes `--tenant-id` flag for `getcred` command
- Adds `--token-url` flag for `getcred` command

VCert Playbook:
- Removes `tenantId` attribute from `config.connection.credentials` object
- Now uses `config.connection.credentials.idP.tokenURL` for VCP service account authentication


marcos-albornoz (approved):

luispresuelVenafi (approved):

BeardedPrincess:
@rvelaVenafi: Seems we updated the codepaths for this, but not the documentation / samples. Should I open an issue, or are you already aware of this?

PR #449, opened 2024-04-04
BUGFIX: cloud toPolicy SAN regex processing

inteon:
The `addStartEndToArray` function was missing for the `p.IpSanRegExs` `p.EmailSanRegExs` `ct.SanRfc822NameRegexes` and `ct.SanUniformResourceIdentifierRegexes` fields.

`addStartEndToArray` also changed the policy because it created an empty slice for a nil input (instead of just returning nil instead).

### DIFF
--- a/pkg/venafi/cloud/certificatePolicies.go
+++ b/pkg/venafi/cloud/certificatePolicies.go
@@ -94,22 +94,31 @@ func (ct certificateTemplate) toPolicy() (p endpoint.Policy) {
 		return s
 	}
 	addStartEndToArray := func(ss []string) []string {
+		// if the array is nil, return nil
+		if ss == nil {
+			return nil
+		}
+
 		a := make([]string, len(ss))
 		for i, s := range ss {
 			a[i] = addStartEnd(s)
 		}
 		return a
 	}
-	if len(ct.SubjectCValues) == 0 {
-		ct.SubjectCValues = []string{".*"}
-	}
+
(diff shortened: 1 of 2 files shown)


pkg/venafi/cloud/certificatePolicies.go
```diff
@@ -94,29 +94,43 @@ func (ct certificateTemplate) toPolicy() (p endpoint.Policy) {
...
+		}
+
 		a := make([]string, len(ss))
 		for i, s := range ss {
 			a[i] = addStartEnd(s)
 		}
 		return a
 	}
-	if len(ct.SubjectCValues) == 0 {
-		ct.SubjectCValues = []string{".*"}
```
inteon, line 104: I do not know why this custom logic was here.
On first sight, this does not seem to be correct behavior.

inteon:
> @inteon can you please rebase your branch?

Done

pkg/venafi/cloud/certificatePolicies.go
inteon: This line was introduced in https://github.com/Venafi/vcert/commit/4429c01ca2ef6a09ab52c29dc4115c0c658bb931.

rvelaVenafi (approved):

marcos-albornoz (approved):

inteon:
I re-signed the commits, since the signature got lost when I clicked the `rebase` button.

PR #450, opened 2024-04-05
Only call DEK endpoints if the CSR is a ServiceGeneratedCSR

inteon:
For the VaaS integration, vcert unconditionally calls the DEK API endpoints.
Similarly to the TPP implementation, we can only call the DEK API calls when a ServiceGeneratedCSR is requested.
This reduces the number of API requests made by vcert.

See the TPP equivalent of this check: https://github.com/Venafi/vcert/blob/21228d0e0de959feddf95acb7d7a22402e2edd23/pkg/venafi/tpp/connector.go#L1359-L1365

### DIFF
--- a/pkg/venafi/cloud/connector.go
+++ b/pkg/venafi/cloud/connector.go
@@ -272,7 +272,7 @@ func (c *Connector) RequestCertificate(req *certificate.Request) (requestID stri
 }
 
 // RetrieveCertificate retrieves the certificate for the specified ID
-func (c *Connector) RetrieveCertificate(req *certificate.Request) (certificates *certificate.PEMCollection, err error) {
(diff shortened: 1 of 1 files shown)


rvelaVenafi:
A little improvement on the code:
```
	if req.CsrOrigin == certificate.ServiceGeneratedCSR || req.FetchPrivateKey {
		var currentId string
		if req.CertID != "" {
			currentId = req.CertID
		} else if certificateId != "" {
			currentId = certificateId
		}

		dekInfo, err := getDekInfo(c, currentId)
		if err != nil  {
		         return nil, err
		}

		req.CertID = currentId
		return retrieveServiceGeneratedCertData(c, req, dekInfo)
	}
```

rvelaVenafi (approved):

EduardoVV (approved):

luispresuelVenafi (approved):

PR #452, opened 2024-04-08
Remove legacy keyPassword from example playbooks

BeardedPrincess:
The property keyPassword was removed from the request section of the playbook several versions ago. However, the sample playbooks still showed them. Since it is just a superfluous property in json, it has been ignored, but it causes plenty of confusion with people using the samples.  This PR removes them.


luispresuelVenafi:
Just to add comment the specific version which this was removed from the `request` object in favor of using it in the task `object` was in version [5.1.0](https://github.com/Venafi/vcert/releases/tag/v5.1.0). We issed to removed. Thank you very much for fixing this @BeardedPrincess

luispresuelVenafi (approved):

marcos-albornoz (approved):

PR #454, opened 2024-04-09
release(5.6.3): Updates CHANGELOG.md for 5.6.3 release

rvelaVenafi:


CHANGELOG.md
```diff
@@ -1,3 +1,28 @@
...
+- Updates all playbook samples, removing deprecated attributes and making sure they work out-of-the-box
+
+VCert SDK:
+- Adds `TokenURL` to `endpoint.Authentication`
+- Cloud Connector will stop using the `TokenURL` attribute from `endpoint.OAuthProvider` and start using the new one 
+(above)
+
+VCert CLI:
+- Internal changes to make use fo the new `TokenURL` attribute
+- Renames `getcred` command flag `--idp-jwt` back to `--external-jwt`
```
sbanic-venafi, line 13: Breaking change. Noting that just going from 5.6.2 to 5.6.3 will break command use for those who are using `--idp-jwt` flag currently.

rvelaVenafi: all previous releases are in a pre-release state and not available to public. So we can do this change safely

rvelaVenafi: I mean: 5.6.0, 5.6.1 and 5.6.2. This new release 5.6.3 is also going into pre-release. So far, average users will get VCert v5.5 when using `go get`

PR #456, opened 2024-04-12
Enables direct supply of a valid accessToken via the playbook for VCP…

BeardedPrincess:
…. also Improves errors on expired access token. (Previously this error was unhandled and failed at parsing an empty response body with an unexpected end of json error.  

Finally, improves playbook validation for VCP credentials, and prevents more than one kind of credential to be set. Like highlander, there can now only be one!  This should keep us from having to make a (wrong) prioritization decision about which credential should be used when multiple are provided.

### DIFF
--- a/pkg/playbook/app/domain/connection.go
+++ b/pkg/playbook/app/domain/connection.go
@@ -104,30 +104,63 @@ func isValidTpp(c Connection) (bool, error) {
 }
 
 func isValidVaaS(c Connection) (bool, error) {
-	// Credentials are not empty
+	// Check if an API key has been provided
 	apikey := false
 	if c.Credentials.APIKey != "" {
 		apikey = true
 	}
 
-	svcaccount := false
+	accesstoken := false
+	if c.Credentials.AccessToken != "" {
+		accesstoken = true
+	}
+
+	// Check if an TokenURL has been provided
+	tokenurl := false
 	if c.Credentials.TokenURL != "" {
+		tokenurl = true
+	}
+
+	// Check if externalJWT has been provided
+	externalJWT := false
+	if c.Credentials.ExternalJWT != "" {
+		externalJWT = true
+	}
+
+	// There's a valid service account IF both externalJWT and tokenURL provided
+	svcaccount := false
+	if externalJWT && tokenurl {
 		svcaccount = true
-	}
-
-	if !apikey && !svcaccount {
-		return false, ErrNoCredentials
-	}
-
-	if apikey {
-		return true, nil
-	}
-
-	if c.Credentials.ExternalJWT == "" {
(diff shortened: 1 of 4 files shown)


pkg/playbook/app/domain/connection.go
```diff
@@ -104,30 +104,63 @@ func isValidTpp(c Connection) (bool, error) {
...
+		return true, nil
 	}
 
-	return true, nil
+	// At this point, there are no valid credentials. Figure out why.
+	if !apikey && !svcaccount && !accesstoken {
+		return false, ErrNoCredentials
+	}
+
+	// If we get here, it's because too many credentials were provided
```
rvelaVenafi, line 162: I dont think we need this check at all. Multiple credentials can be passed to a Connector and they will be used in priority. For VCP  it is:
1. access token, then
2. service account, and then
3. api key

BeardedPrincess: it may be overkill for sure, but this is only in the playbook. I figured it would be worthwhile to force the playbook user to be specific, and not specify multiple. It doesn't fix operations for sure, but it clears up confusion. Happy to exclude it, but I don't think it hurts either.

rvelaVenafi (approved):

EduardoVV (approved):

luispresuelVenafi (approved):

pkg/playbook/app/domain/connection.go
```diff
@@ -104,30 +104,63 @@ func isValidTpp(c Connection) (bool, error) {
...
+		// JWT Provided without token url
+		return false, ErrNoVCPTokenURL
+	} else if tokenurl && !externalJWT {
+		// Tokenurl without an external JWT
+		return false, ErrNoExternalJWT
 	}
 
-	if !apikey && !svcaccount {
-		return false, ErrNoCredentials
+	// If there's just an API key provided, that's good
```
sbanic-venafi, line 142: Your logic here is a bit too complex.
I think what you're trying for here (after line 142) is:
```
if !svcaccount && !apikey && !accesstoken {
	// none of the credential options were provided
	return false, ErrNoCredentials
}
// if we got here then at least one of the credential options was provided
if (svcaccount && apikey) || (svcaccount && accesstoken) || (apikey && accesstoken) {
	// more than one credential option is not acceptable
	return false, ErrAmbiguousVCPCreds
}
// if we got here then only one credential option was provided (which is what we want)
return true, nil
```

sbanic-venafi (changes requested):

pkg/playbook/app/domain/connection.go
BeardedPrincess: much better, thanks

PR #475, opened 2024-05-13
Improves keystore querying and provisioning mutation with new refactoring

luispresuelVenafi:


rvelaVenafi (approved):

pkg/webclient/cloudproviders/cloudproviders.go
```diff
@@ -50,3 +51,61 @@ func (c *CloudProvidersClient) GetCloudProviderByName(ctx context.Context, name
 		KeystoresCount: cp.GetKeystoresCount(),
 	}, nil
 }
+
+func (c *CloudProvidersClient) GetCloudKeystores(ctx context.Context, cloudKeystoreID *string, cloudKeystoreName *string, cloudProviderID *string, cloudProviderName *string) (*domain.CloudKeystore, error) {
+
+	if cloudKeystoreID == nil {
+		if cloudKeystoreName == nil || (cloudProviderID == nil && cloudProviderName == nil) {
+			return nil, fmt.Errorf("any of keystore ID or both (any of Provider Name of Provider ID) and Keystore Name must be provided for provisioning")
```
sbanic-venafi, line 59: nitpick: I think this is more legible:
```
following are accepted for provisioning: keystore ID, or both keystore Name and provider Name, or both keystore Name and provider ID
```

pkg/webclient/cloudproviders/cloudproviders.go
```diff
@@ -50,3 +51,61 @@ func (c *CloudProvidersClient) GetCloudProviderByName(ctx context.Context, name
 		KeystoresCount: cp.GetKeystoresCount(),
 	}, nil
 }
+
+func (c *CloudProvidersClient) GetCloudKeystores(ctx context.Context, cloudKeystoreID *string, cloudKeystoreName *string, cloudProviderID *string, cloudProviderName *string) (*domain.CloudKeystore, error) {
```
sbanic-venafi, line 55: Did you mean `GetCloudKeystore` (no `s`)?

luispresuelVenafi: Good catch. I was still thinking of previous function that fetched them all, but yeah, in there I'm already reducing the amount to 1

pkg/webclient/cloudproviders/cloudproviders.go
luispresuelVenafi: done by latest commit

pkg/webclient/cloudproviders/cloudproviders.go
luispresuelVenafi: Fixed by latest commit

sbanic-venafi (approved):

pkg/webclient/cloudproviders/cloudproviders.go
```diff
@@ -50,3 +51,61 @@ func (c *CloudProvidersClient) GetCloudProviderByName(ctx context.Context, name
 		KeystoresCount: cp.GetKeystoresCount(),
 	}, nil
 }
+
+func (c *CloudProvidersClient) GetCloudKeystore(ctx context.Context, cloudKeystoreID *string, cloudKeystoreName *string, cloudProviderID *string, cloudProviderName *string) (*domain.CloudKeystore, error) {
+
+	if cloudKeystoreID == nil {
+		if cloudKeystoreName == nil || (cloudProviderID == nil && cloudProviderName == nil) {
+			return nil, fmt.Errorf("following are accepted for provisioning: keystore ID, or both keystore Name and provider Name, or both keystore Name and provider ID")
```
rvelaVenafi, line 59: While the three combinations are possible, users don't need to know that a combination of ID and name is available. 
If the user can get the name of the keystone, they can also get the provider's name.

In that sense, we should only advise two scenarios:
- keystore ID or
- provider name and keystore name

rvelaVenafi (changes requested):

pkg/webclient/cloudproviders/cloudproviders.go
luispresuelVenafi: I mean, to that sense in the webclient, the endpoint provides the ability to set the Cloud Provider ID. I was not constraining it in there since I was already doing so in the Provisioning interface method. Do you think it also applies for in there?

rvelaVenafi: In that case, constrain it. They dont need all options available. Either keystore ID or provider/keystore names

luispresuelVenafi: Done by latest changes

rvelaVenafi (approved):

pkg/venafi/cloud/connector.go
```diff
@@ -792,33 +792,19 @@ func (c *Connector) ProvisionCertificate(req *endpoint.ProvisioningRequest, opti
...
-
-	if len(data.CloudKeystores.Nodes) != 1 {
-		return nil, fmt.Errorf("could not find keystore with provided information. KeystoreID: %s, KeystoreName: %s, ProviderName: %s", keystoreIDInput, keystoreNameInput, providerNameInput)
-	}
-
-	// grabbing found keystore for later
-	cloudKeystore := data.CloudKeystores.Nodes[0]
-
-	keystoreIDString := cloudKeystore.GetId()
+	keystoreIDString := cloudKeystore.ID
```
marcos-albornoz, line 807: I'm wondering how the cloudKeystore.GetId() existed and why now no... it's always preferable to use an access method instead of refer directly to the property of an struct

marcos-albornoz (approved):

pkg/venafi/cloud/connector.go
luispresuelVenafi: because the previous value was directly from the generated code. Now I'm using a web client, and currently the struct I'm using doesn't have those methods defined. I didn't see them necessary

PR #476, opened 2024-05-14
vcert playbook - fixes issue with timeout

luispresuelVenafi:
vcert playbook - fixes issue when default time is not set, it was taking 30 secs instead of 180, also added chance to set own timeout

### DIFF
--- a/pkg/playbook/app/vcertutil/vcertutil.go
+++ b/pkg/playbook/app/vcertutil/vcertutil.go
@@ -23,7 +23,6 @@ import (
 	"crypto/x509/pkix"
 	"errors"
 	"fmt"
-	"github.com/Venafi/vcert/v5/pkg/verror"
 	"log"
 	"net/http"
 	"strings"
@@ -37,6 +36,7 @@ import (
 	"github.com/Venafi/vcert/v5/pkg/playbook/app/domain"
 	"github.com/Venafi/vcert/v5/pkg/util"
 	"github.com/Venafi/vcert/v5/pkg/venafi/tpp"
+	"github.com/Venafi/vcert/v5/pkg/verror"
 )
 
 // EnrollCertificate takes a Request object and requests a certificate to the Venafi platform defined by config.
--- a/pkg/playbook/app/vcertutil/vcertutil.go
+++ b/pkg/playbook/app/vcertutil/vcertutil.go
@@ -23,7 +23,6 @@ import (
 	"crypto/x509/pkix"
 	"errors"
 	"fmt"
-	"log"
 	"net/http"
 	"strings"
 	"time"
@@ -105,7 +104,7 @@ func buildClient(config domain.Config, zone string, timeout int) (endpoint.Conne
 	var connectionTrustBundle *x509.CertPool
 
 	if vcertConfig.ConnectionTrust != "" {
-		log.Println("Using trust bundle in custom http client")
+		zap.L().Debug("Using trust bundle in custom http client")
 		connectionTrustBundle = x509.NewCertPool()
 		if !connectionTrustBundle.AppendCertsFromPEM([]byte(vcertConfig.ConnectionTrust)) {
 			return nil, fmt.Errorf("%w: failed to parse PEM trust bundle", verror.UserDataError)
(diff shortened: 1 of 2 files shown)


pkg/playbook/app/vcertutil/vcertutil.go
```diff
@@ -91,6 +96,27 @@ func buildClient(config domain.Config, zone string) (endpoint.Connector, error)
...
+	vcertConfig.Client = &http.Client{
+		Timeout: time.Duration(DefaultTimeout) * time.Second,
+	}
+	if timeout > 0 {
+		vcertConfig.Client.Timeout = time.Duration(timeout) * time.Second
+	}
+	var connectionTrustBundle *x509.CertPool
+
+	if vcertConfig.ConnectionTrust != "" {
+		log.Println("Using trust bundle in custom http client")
```
rvelaVenafi, line 108: use zap.log 
You need to pass it to the function

rvelaVenafi (changes requested):

sbanic-venafi (approved):

pkg/playbook/app/vcertutil/vcertutil.go
```diff
@@ -91,6 +95,28 @@ func buildClient(config domain.Config, zone string) (endpoint.Connector, error)
...
+	vcertConfig.Client = &http.Client{
+		Timeout: time.Duration(DefaultTimeout) * time.Second,
+	}
+	if timeout > 0 {
+		vcertConfig.Client.Timeout = time.Duration(timeout) * time.Second
+	}
+	var connectionTrustBundle *x509.CertPool
+
+	if vcertConfig.ConnectionTrust != "" {
+		zap.L().Debug("Using trust bundle in custom http client")
```
rvelaVenafi, line 107: Don't just invoke zap, pass the actual logger used by the program. This way zap will create a default logger without the customizations from the plyabook

rvelaVenafi (approved):

pkg/playbook/app/vcertutil/vcertutil.go
rvelaVenafi, line 107: Don't just invoke zap, pass the actual logger used by the program. This way zap will create a default logger without the customizations from the plyabook

rvelaVenafi (changes requested):

BeardedPrincess (approved):
This looks good.. are there more changes coming for this? (I only see the one file changed, I'm assuming we'll need to update

pkg/playbook/app/vcertutil/vcertutil.go
BeardedPrincess, line 1: @luispresuelVenafi 

Line 76 (in the 'else' statement of if supports synchronous response) is for some reason also setting the timeout on vrequest before taking action. I'm not sure that causes any issue, but it either needs to be updated to use the new timeout variable, or removed if it has no effect. (client is already created, so not sure if it actually does anything)

pkg/playbook/app/vcertutil/vcertutil.go
```diff
@@ -69,7 +73,6 @@ func EnrollCertificate(config domain.Config, request domain.PlaybookRequest) (*c
 		zap.L().Debug("successfully requested certificate", zap.String("requestID", reqID))
 
 		vRequest.PickupID = reqID
-		vRequest.Timeout = 180 * time.Second
```
rvelaVenafi, line 72: If you remove this line, and the first call to retrieve fails (certificate is not in the ISSUED state), you will have to keep invoking the retrieve function or fail. The timeout in the request object is to loop over the certificate retrieval for slow CAs.

rvelaVenafi: Do not remove it. It makes sense.

rvelaVenafi (changes requested):

pkg/playbook/app/vcertutil/vcertutil.go
luispresuelVenafi: It doesn't make sense, that timeout is already set before

rvelaVenafi (approved):

PR #481, opened 2024-05-17
Update powershell.go

sabixx:
Remove semicolon from PowerShell execution command to fix ambiguous error messages, e.g., caused when running into executionpolicy issues.


sabixx:
I was wrong, this leads to issues.

PR #485, opened 2024-05-23
feat(cloud-provisioning): Adds function to retrieve machine identity by ID

rvelaVenafi:
Moves cloud providers graphql API to the webclient folder

### DIFF
--- a/pkg/domain/cloudproviders.go
+++ b/pkg/domain/cloudproviders.go
@@ -1,5 +1,7 @@
 package domain
 
+import "github.com/google/uuid"
+
 type CloudProvider struct {
 	ID             string
 	Name           string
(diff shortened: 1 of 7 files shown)


EduardoVV (approved):

pkg/webclient/cloudproviders/genqlient.graphql
```diff
@@ -33,3 +33,32 @@ query GetCloudProviderByName($name: String!){
...
+                    gcpId
+                    name
+                }
+            }
+            status
+            statusDetails
+            certificateId
+        }
+    }
+}
```
luispresuelVenafi, line 64: Nit: could you add an end of line here?

pkg/webclient/cloudproviders/genqlient.yaml
```diff
@@ -1,6 +1,6 @@
 # For full documentation see:
 # https://github.com/Khan/genqlient/blob/main/docs/genqlient.yaml
-schema: ../../../api/graphql/schema.graphql
```
luispresuelVenafi, line 3: Isn't a standard in graphql to have the file where it was before this change?

rvelaVenafi: Yes, if it were an API for a service. In this case, the API is to generate a web client. It does not make sense to be in a different folder than the client itself.

pkg/webclient/cloudproviders/genqlient.graphql
rvelaVenafi: Done

luispresuelVenafi (approved):

pkg/domain/cloudproviders.go
```diff
@@ -27,3 +83,112 @@ type GetCloudKeystoreRequest struct {
...
+
+func (ccm *CertificateCloudMetadata) GetType() string {
+	typ := ccm.GetValue("__typename")
+	if typ == nil {
+		return CloudMetadataUnknown
+	}
+	switch typ {
+	case "AWSCertificateMetadata":
+		return CloudMetadataACM
+	case "AzureCertificateMetadata":
```
sbanic-venafi, line 153: These (GCM, Azure) are wrong. Reversed.

pkg/domain/cloudproviders.go
```diff
@@ -27,3 +83,112 @@ type GetCloudKeystoreRequest struct {
...
+	CertificateID     uuid.UUID
+	Metadata          *CertificateCloudMetadata
+	Status            MachineIdentityStatus
+	StatusDetails     string
+}
+
+type GetCloudMachineIdentityRequest struct {
+	KeystoreID        *string
+	MachineIdentityID *string
+	Fingerprints      []string
```
sbanic-venafi, line 191: If MI ID is mandatory, then why multiple fingerprints? Just curious, it seems odd.

rvelaVenafi: This is just a representation of the filters used by machineIdentities query filters.
Right now we don't need to query for multiple MI, but in the future, this request could be used to query for both a single MI or multiple MIs (like by keystoreID or by fingerprints)

pkg/domain/cloudproviders.go
rvelaVenafi: Thanks. Good catch! Fixed by latest commit

PR #509, opened 2024-08-09
remove store validation

ipitythebug:
Recommend removing store validation - use cases may vary from one enterprise to the next. If this can't be removed - I'd recommend adding WebHosting as it is designed to support SNI use cases

### DIFF
--- a/pkg/playbook/app/domain/installation.go
+++ b/pkg/playbook/app/domain/installation.go
@@ -32,8 +32,10 @@ const (
 	capiLocationLocalMachine = "localmachine"
 )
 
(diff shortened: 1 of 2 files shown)


BeardedPrincess (approved):

Pmaraveyias:
Would it be better to make store validation optional? I'm happy to contribute the code if there's value in keeping the option.

R3LentL355:
> Would it be better to make store validation optional? I'm happy to contribute the code if there's value in keeping the option.

That would be helpful - I think that any sort of allow/deny "policy" such as this should be something we would consider as being user-configurable.

Pmaraveyias:
Superceded by https://github.com/Venafi/vcert/pull/524

PR #512, opened 2024-08-19
Set log prefix in main.go instead of endpoint.go

inteon:
vCert sets a log prefix on the global logger in endpoint.go.
This unexpectedly alters the logger when importing vCert.
In order to not modify the global logger when importing vCert as a go library, I moved the `SetPrefix` call to `main.go` which is only used in the vCert binary but when importing vCert as a library.

### DIFF
--- a/cmd/vcert/main.go
+++ b/cmd/vcert/main.go
@@ -35,6 +35,10 @@ var (
 	exit   = os.Exit
 )
 
+func init() {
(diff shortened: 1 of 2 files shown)


BeardedPrincess (approved):

maelvls:
Hey. Would you be able to give a second review @luispresuelVenafi? Thanks

luispresuelVenafi (approved):

PR #513, opened 2024-08-20
Add examples for running playbooks with ansible while using TLSPC SVC accounts

vertigo-one:
Adding Sample Ansible playbooks that can be used to trigger [VCert Playbooks](https://github.com/Venafi/vcert/blob/master/README-PLAYBOOK.md) using Venafi Control Plane (VCP) [Service Accounts](https://docs.venafi.cloud/vcs-platform/serviceaccounts/about-service-accounts/). 

This method of execution allows for TLSPC svc accounts to be used without storing the credentials to authentication to the IDP on each VM, or setting up scheduled tasks on each managed VM. 

Contains:
- New Examples Folder for Ansible.
- folder README.md file explaining use case and requirements.
- Ansible playbook examples for Azure Entra, ADFS, and Auth0. 
- Example vcert playbook file.

### DIFF
--- a/examples/ansible/testvcertplaybook.yaml
+++ b/examples/ansible/testvcertplaybook.yaml
@@ -1,0 +1,20 @@
+config:
+  connection:
+    platform: vaas
+    credentials:
+      tokenURL: <Your Service Account Token URL - Copy from VCP> 
+      externalJWT: '{{ Env "TLSPC_EXTERNAL_JWT" }}'
+certificateTasks:
+  - name: myCertificate
+    renewBefore: 31d
+    request:
+      csr: service
+      subject:
+        commonName: "ansibletest.lab.corp.net"
+      zone: "<myapp>\\<issuingtemplate>"
+    installations:
+      - format: PEM
+        file: "./cert.cer"
+        chainFile: "./chain.cer"
+        keyFile: "./key.pem" 
+        afterInstallAction: "echo Success!!!"
--- a/examples/ansible/testvcertplaybook.yaml
+++ b/examples/ansible/testvcertplaybook.yaml
@@ -5,16 +5,24 @@ config:
       tokenURL: <Your Service Account Token URL - Copy from VCP> 
       externalJWT: '{{ Env "TLSPC_EXTERNAL_JWT" }}'
 certificateTasks:
-  - name: myCertificate
+  - name: appcert
+    setEnvVars: #delete the environment variables that you do not need.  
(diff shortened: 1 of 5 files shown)


examples/ansible/testvcertplaybook.yaml
```diff
@@ -0,0 +1,20 @@
...
+  connection:
+    platform: vaas
+    credentials:
+      tokenURL: <Your Service Account Token URL - Copy from VCP> 
+      externalJWT: '{{ Env "TLSPC_EXTERNAL_JWT" }}'
+certificateTasks:
+  - name: myCertificate
+    renewBefore: 31d
+    request:
+      csr: service
```
BeardedPrincess, line 11: @vertigo-one : Would you mind changing to "local" instead of "service"? I'm trying to encourage a best practice of generating the private keys on the machines, rather than in the platform. (There's no need, it's more secure, and for TLSPC, doesn't require that we have a vSat!)

examples/ansible/testvcertplaybook.yaml
```diff
@@ -0,0 +1,20 @@
...
+      subject:
+        commonName: "ansibletest.lab.corp.net"
```
BeardedPrincess, line 13: We probably would want to use the variables here to determine the CN based on hostname?

examples/ansible/testvcertplaybook.yaml
vertigo-one: Sure. Not a problem.

examples/ansible/testvcertplaybook.yaml
vertigo-one: Agreed on the DNS SAN. I'll add that. 

I was not putting in any variables or expressions to determine the CN by the hostname as the usecase this was made for was not server certs but applications hosted by the server, which have differing FQDNs they are accessible by than the server hostname. Along the lines of  `myapp.corp.net` instead of `WKS115PKM.corp.net`

examples/ansible/testvcertplaybook.yaml
vertigo-one: Updated in bf95fa38a9b528b083446a14790d1543abd39f06

examples/ansible/testvcertplaybook.yaml
vertigo-one: Updated in bf95fa38a9b528b083446a14790d1543abd39f06

JigarAtVenafi (approved):

PR #516, opened 2024-09-05
Add SetAuthentication function to authenticate 'passively'

inteon:
When using vcert as a library, being able to create a vcert client without sending a self-check request to the server can help reduce unnecessary API calls.
It gives the caller more control over when the self-checks occur.
Eg. when creating clients dynamically based on cached credentials.

### DIFF
--- a/pkg/venafi/cloud/connector.go
+++ b/pkg/venafi/cloud/connector.go
@@ -91,7 +91,6 @@ type Connector struct {
 	apiKey                string
 	accessToken           string
 	verbose               bool
-	user                  *userDetails
 	trust                 *x509.CertPool
 	zone                  cloudZone
 	client                *http.Client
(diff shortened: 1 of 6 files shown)


hawksight:
@luispresuelVenafi / @rvelaVenafi / @marcos-albornoz / @EduardoVV - Any chance you might be able to review this PR from @inteon?

pkg/venafi/tpp/connector.go
```diff
@@ -45,7 +45,6 @@ type Connector struct {
 	apiKey      string
 	accessToken string
 	verbose     bool
-	Identity    identity
```
inteon, line 48: `Identity` was not used anywhere.

pkg/venafi/cloud/connector.go
```diff
@@ -91,7 +91,6 @@ type Connector struct {
 	apiKey                string
 	accessToken           string
 	verbose               bool
-	user                  *userDetails
```
inteon, line 94: `user` was not meaningfully used anywhere.

pkg/venafi/cloud/connector.go
```diff
@@ -135,35 +134,45 @@ func (c *Connector) Ping() (err error) {
...
 		tokenResponse, err := c.GetAccessToken(auth)
 		if err != nil {
 			return err
 		}
 		c.accessToken = tokenResponse.AccessToken
 	} else if auth.APIKey != "" {
-		// 3. API key. Get user to test authentication
+		// 3. API key. Assign it to connector
 		c.apiKey = auth.APIKey
-		url := c.getURL(urlResourceUserAccounts)
```
inteon, line 157: This check has been moved to the `getUserDetails` function which is called by `Authenticate`.

pkg/venafi/tpp/connector.go
```diff
@@ -143,13 +156,6 @@ func (c *Connector) Authenticate(auth *endpoint.Authentication) (err error) {
 
 		resp := result.(authorizeResponse)
 		c.apiKey = resp.APIKey
-
-		if c.client != nil {
-			c.Identity, err = c.retrieveSelfIdentity()
```
inteon, line 148: These calls have been moved to the `Authenticate` function.

PR #527, opened 2024-09-16
Adds capability to define EKUs to be included in CSRs

BeardedPrincess:
In some instances, it may be necessary to include desired EKUs within the CSR when vCert is responsible for generating the private key and subsequent CSR.  The vast majority of CAs will ignore this, and issue certificates with properties governed by the CA.  However, we have recently encountered one instance where the upstream CA requires for these to be included within the CSR. 

This PR primarily adds a new hidden cmdLine flag to the CLI (`--eku`) which can be used during enroll or gencsr to include additional EKUs as defined.  This is limited to currently known and supported EKUs in the crypto/x509 library.  This functionality is also extended to the playbook use-case as well, by introducing a new YAML field (`eKU`) within the request section of a certificate task.

Furthermore, this PR exposes the `crypto/x509.CertificateRequest.ExtraAttributes` property through the `vcert/certificate.Request` object which will enable developers using  vCert SDK to add additional attributes to the CSR during the request process enabling use-cases that would otherwise require changes directly to the SDK.  This was done because the current Attributes property is being deprecated.


Makefile
```diff
@@ -48,6 +48,9 @@ build: get
 	env GOOS=linux   GOARCH=arm64 go build $(GO_LDFLAGS) -o bin/linux/vcert_arm       ./cmd/vcert
 	env GOOS=linux   GOARCH=amd64 go build $(GO_LDFLAGS) -o bin/linux/vcert           ./cmd/vcert
 	env GOOS=linux   GOARCH=386   go build $(GO_LDFLAGS) -o bin/linux/vcert86         ./cmd/vcert
+	env GOOS=linux   GOARCH=arm GOARM=5 go build $(GO_LDFLAGS) -o bin/linux/vcert_arm32v5       ./cmd/vcert
```
marcos-albornoz, line 51: Is it possible that you do these changes in a separated PR?

pkg/certificate/extKeyUsage.go
```diff
@@ -0,0 +1,400 @@
+/*
```
marcos-albornoz, line 1: Do you have the unit tests for these changes?

Makefile
luispresuelVenafi: This definitely should go in another MR, since is not mentioned in the current description. Adding support to new type of binaries, will increase the MR dependency in another changes, as we need to internal changes in our pipeline along with this changes to be able to actually deploy these new binaries, when another VCert version that contains this changes is released.

cmd/vcert/flags.go
```diff
@@ -239,6 +239,13 @@ var (
 		Hidden: true,
 	}
 
+	flagExtKeyUsage = &cli.StringSliceFlag{
+		Name: "eku",
+		Usage: "Use to specify an Extended Key Usage type to be included in the generated CSR. Common options include ServerAuth & ClientAuth. " +
+			"This option can be repeated to specify more than one value like this: --eku ServerAuth --eku ClientAuth etc. ",
+		Hidden: true,
```
angelmoo, line 246: do we know why this will be hidden?

pkg/certificate/certificate_test.go
```diff
@@ -56,6 +58,9 @@ func getCertificateRequestForTest() *Request {
 
 	}
 	req.IPAddresses = ips
+
+	req.ExtKeyUsages = *NewExtKeyUsageSlice("ClientAuth")
```
angelmoo, line 62: nit: may be a test with ServerAuth could be added, but not mandatory. just to have a different case.

angelmoo (approved):

pkg/certificate/certificate_test.go
marcos-albornoz: In fact that was a mistake... the ExtKeyUsage in the slice must have been "ServerAuth"

marcos-albornoz: Fixed now

marcos-albornoz (approved):

PR #532, opened 2024-09-23
Proxy client fix for VCert Playbook and more

luispresuelVenafi:
- Adds retry logic for waiting to get certificate in case is not ready at the certificate inventory. 
- Refactors code logic.
- Adds missing proxy config when creating http client


pkg/venafi/cloud/connector.go
```diff
@@ -1274,34 +1267,55 @@ type managedCertificate struct {
...
+						if time.Now().After(startTime.Add(timeout)) {
+							return nil, endpoint.ErrRetrieveCertificateTimeout{CertificateID: certificateId}
+						}
+					} else {
+						respError := fmt.Sprintf("unexpected status code on Venafi Cloud certificate search. Status: %d\n", statusCode)
+						for _, e := range respErrors {
+							respError += fmt.Sprintf("Error Code: %d Error: %s\n", e.Code, e.Message)
+						}
+						return nil, errors.New(respError)
+					}
```
marcos-albornoz, line 1307: It's seeming this block of code is doing very similar tasks to the block of code in line 1310 - 1315 with slightly difference in the last part which I consider should be equals in both cases

marcos-albornoz (approved):

EduardoVV (approved):

PR #534, opened 2024-10-01
Update README-PLAYBOOK.md

sabixx:
Added RSA 3072, as it's also a supported key size.


luispresuelVenafi (approved):
Approving as this is true since, unlike TPP only side in VCert PKG, we don't have a restriction on our Playbook side, rather we let the restriction to be handled by the platform

luispresuelVenafi:
@sabixx could you update this PR with the latest changes in VCert?

PR #551, opened 2025-03-05
Support TPP v25.1 and higher

stkomitov:


BeardedPrincess:
I this effort, it would be great if we could also merge or incorporate the changes from #527.  I imagine the work to integrate the PKIX support will impact the certificate signing request modules which that PR had to tackle since the original/current code-base adds a couple of layers of abstraction from the crypto/x509 library.

stkomitov:
> I this effort, it would be great if we could also merge or incorporate the changes from #527. I imagine the work to integrate the PKIX support will impact the certificate signing request modules which that PR had to tackle since the original/current code-base adds a couple of layers of abstraction from the crypto/x509 library.

Most of the TPP WebSDK API calls are backward compatible, so in my PR I would mostly handle the Policy Get/Set commands. The certificate API (request, renew, check policy...) is completely backward compatible with TPP < 25.1. So, your changes doesn't overlap with the changes in this PR, but our team will work on merging #527 and release it in the same vcert version, hopefully by the end of this week.
The full PKIX support and support for PQ algorithms would be introduced later.

marcos-albornoz:
Should it not be considered to do some update to include the `PkixParameterSet` as part of the `PolicySpecification` returned by the `venafi.tpp.Connector.GetPolicy()`?

stkomitov:
> Should it not be considered to do some update to include the `PkixParameterSet` as part of the `PolicySpecification` returned by the `venafi.tpp.Connector.GetPolicy()`?

I added it because it is cleaner when we need to build the TPP policy in `func BuildTppPolicy(ps *PolicySpecification) TppPolicy`. Later we would need to update vcert to also store/read the `PkixParameterSet` attribute to/from files as well (policy as json/yaml), because TPP would only work with that attribute in later versions.

luispresuelVenafi (approved):

EduardoVV (approved):

PR #557, opened 2025-04-10
fix: Migrates from deprecated library ioutil

luispresuelVenafi:
- Migrates from deprecated library `ioutil`
- Handles errors that missed handling

### DIFF
--- a/examples/server/main.go
+++ b/examples/server/main.go
@@ -2,7 +2,7 @@ package main
 
 import (
 	"fmt"
-	"io/ioutil"
+	"io"
 	"log"
 	"net/http"
 	"os"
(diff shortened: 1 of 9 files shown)


examples/server/main.go
```diff
@@ -52,7 +52,11 @@ func initConfig() *vcert.Config {
 	}
 	trustBundleFilePath := os.Getenv("TRUST_BUNDLE_PATH")
 	if trustBundleFilePath != "" {
-		buf, err := ioutil.ReadFile(trustBundleFilePath)
+		file, err := os.Open(trustBundleFilePath)
+		if err != nil {
+			panic(err)
+		}
+		buf, err := io.ReadAll(file)
```
stkomitov, line 59: I think we also need to call Close in `defer`:
```go
defer func() {
    err = file.Close()
}()
```
On the other places where we call .Open as well.

PR #666, opened 2026-06-11
NGTS: use api.strata instead of api.sase

maelvls:
As mentioned in https://github.com/Venafi/vcert/pull/634#discussion_r3287630328, I propose to update the default API URL's domain from `api.sase.paloaltonetworks.com` to `api.strata.paloaltonetworks.com`.

## Why

Looking at the [pan.dev docs](https://github.com/PaloAltoNetworks/pan.dev/blob/eb109fb12d4a9861b4854a0a3f56e5bc7682ad11/products/scm/docs/release-notes/november2024.md#new-api-platform-fqdn):

> The FQDN for all Strata Cloud Manager APIs has been updated to reflect our broader platform capabilities. The new FQDN is api.strata.paloaltonetworks.com. The original FQDN of api.sase.paloaltonetworks.com will continue to work for the time being. However, **all API documentation, tooling, SDKs, and other materials will be updated to reflect the new FQDN.**

Related:

- https://github.com/cert-manager/cert-manager/pull/8779#discussion_r3287614903


simeoncybr (approved):

stkomitov (approved):

pkg/venafi/ngts/connector.go
```diff
@@ -53,7 +53,7 @@ import (
 type urlResource string
 
 const (
-	apiURL                                        = "api.sase.paloaltonetworks.com/ngts"
+	apiURL                                        = "api.strata.paloaltonetworks.com/ngts"
```
TomaszSwierszczCA, line 56: How about dev version? Will be changed too?

maelvls: No, I just tested and it looks like `dev.api.strata.paloaltonetworks.com` doesn't exist. Weird though. I've adjusted my commit to only use `api.strata` in prod URLs and keep `api.sase` in dev.

```
$ nslookup dev.api.strata.paloaltonetworks.com
Server:         10.220.30.1
Address:        10.220.30.1#53

** server can't find dev.api.strata.paloaltonetworks.com: NXDOMAIN
```

config_test.go
```diff
@@ -74,7 +74,7 @@ tpp_user = admin
 cloud_zone = Default`
 
 const validNgtsConfig1 = `
-url = https://xyz.api.sase.paloaltonetworks.com/ngts
+url = https://xyz.api.strata.paloaltonetworks.com/ngts
```
maelvls: Not sure what these should be, but looking at the token URL, this "example" URL is meant to be the "dev" URL. So I'll revert back to `api.sase`.

pkg/venafi/ngts/connector.go
maelvls: I've asked this in Slack: https://panw-rnd.slack.com/archives/C0998AVC5K2/p1781537484912219

> api.sase vs. api.strata in dev?
> 
> Hey. I've read in the official [pan.dev](http://pan.dev/) docs that all new integrations should start using api.strata instead of api.sase. But what about the dev environment? Right now, the record [dev.api.strata.paloaltonetworks.com](http://dev.api.strata.paloaltonetworks.com/) doesn't exist.

config_test.go
```diff
@@ -74,7 +74,7 @@ tpp_user = admin
 cloud_zone = Default`
 
 const validNgtsConfig1 = `
-url = https://xyz.api.sase.paloaltonetworks.com/ngts/ngts
+url = https://xyz.api.sase.paloaltonetworks.com/ngts
```
maelvls, line 77: Oops. Thanks for fixing this!

PR #682, opened 2026-08-04
fix: read TPP's PkixParameterSet key-algorithm policy (25.1+)

wallrj-cyberark:
## Problem

From TPP 25.1 onwards, the `Certificates/CheckPolicy` endpoint no longer locks the deprecated `KeyPair.KeyAlgorithm`/`KeySize`/`EllipticCurve` fields when a policy folder's allowed key algorithms are configured via the newer AlgorithmSelector API. Instead, the allowed algorithms are expressed as a list of PKIX OIDs in `KeyPair.PkixParameterSet.Values`.

vcert has **two independent consumers** of this same `CheckPolicy` response, both of which only read the deprecated fields:
- `pkg/venafi/tpp/tpp.go`'s `serverPolicy.toPolicy()`, used by `ReadPolicyConfiguration`/`ReadZoneConfiguration` to populate `endpoint.Policy.AllowedKeyConfigurations` — this is what a calling application (e.g. cert-manager's `approver-policy` Venafi plugin) uses to actually enforce policy.
- `pkg/policy/policyUtils.go`'s `BuildPolicySpecificationForTPP`, used by the `vcert getpolicy`/`setpolicy` CLI commands to report/manage policy.

Both silently treated any TPP 25.1+ policy folder using the new PKIX mechanism as unrestricted: `endpoint.Policy.AllowedKeyConfigurations` ended up containing every supported RSA size and ECDSA curve, and `vcert getpolicy` reported no key type/size restriction under `policy.keyPair` at all — regardless of what the TPP administrator had actually configured.

## Fix

- Adds a `PkixParameterSet` field to `serverPolicy.KeyPair` (`pkg/venafi/tpp/tpp.go`) and to `KeyPairResponse` (`pkg/policy/policyStructures.go`) to capture `KeyPair.PkixParameterSet.Values` from the `CheckPolicy` response in both code paths.
- Adds a shared reverse lookup table, `policy.PkixToKeyAlgorithms` (`pkg/policy/policyUtils.go`), mapping the PKIX OIDs (Venafi's IANA private enterprise arc, `1.3.6.1.4.1.28783...`) back to key type/size/curve — the inverse of the existing forward mapping `KeyAlgorithmsToPKIX`, which is also extended to cover RSA 8192 and ED25519 (both previously missing). ED25519 (`1.3.6.1.4.1.28783.10.3.25519`) sits on its own arc rather than under the ECDSA one; it is keyed under `ECC` so that it decodes to a curve name `certificate.EllipticCurve.Set` understands and so that `toPolicy` emits the ECDSA `AllowedKeyConfiguration` that `endpoint.checkKey` matches an ED25519 key against. `ED25519` is added to `TppEllipticCurves` for the same reason RSA 8192 was, so that `getpolicy` output remains valid `setpolicy` input.
- `serverPolicy.toPolicy()` and `BuildPolicySpecificationForTPP` both now prefer `KeyPair.PkixParameterSet.Values` when locked, falling back to the deprecated `KeyAlgorithm`/`KeySize`/`EllipticCurve` fields for TPP versions before 25.1.
- Adds unit test coverage for both paths: `TestConvertServerPolicyToInternalPolicy_PkixParameterSet` (single/multiple RSA sizes, a single ECDSA curve, mixed RSA+ECDSA, the legacy fallback, an unrecognised-OID error) plus an ED25519 subtest) and `TestBuildPolicySpecificationForTPPPkixParameterSet`/`...Ed25519`/`...UnrecognisedOID`. `TestPkixToKeyAlgorithmsIsInverseOfKeyAlgorithmsToPKIX` now also asserts that every curve in the table is accepted by validation, mirroring the existing RSA size assertion — that check is what catches an OID being added without its `TppEllipticCurves` entry.

**ED25519 caveat.** This makes vcert read and report an ED25519 policy correctly. It does **not** on its own make an ED25519 request succeed: `pkg/endpoint/endpoint.go` asserts the parsed CSR public key to `*ed25519.PublicKey`, but Go's `x509.ParseCertificateRequest` returns `ed25519.PublicKey` by value, so every ED25519 CSR is rejected with `invalid key in csr`. That is a separate one-character defect and is deliberately left out of this PR.

## Test evidence

**Enforcement path** (`serverPolicy.toPolicy()`): verified end-to-end against a live TPP 25.3 instance using our internal cert-manager `approver-policy` Venafi plugin's integration test suite, which calls this exact code path (`ReadPolicyConfiguration`/`ReadZoneConfiguration` → `toPolicy()`) to enforce key-type/size policy on CertificateRequests. Three integration subtests exercise this scenario against a policy folder locked to RSA 4096/8192 via the new PKIX mechanism — submitting an ECDSA key, an unsupported RSA size, and an RSA key below the locked minimum. Before this fix vcert silently allowed all three; consuming this branch via a `go.mod replace`, all three now correctly deny the request.

**CLI path** (`BuildPolicySpecificationForTPP`): verified live with `vcert getpolicy` against the same TPP 25.3 policy folder (locked to RSA 4096/8192 via PkixParameterSet). Before this fix:

```json
"policy": { "keyPair": { "reuseAllowed": true } }
```

After this fix:

```json
"policy": { "keyPair": { "keyTypes": ["RSA"], "rsaKeySizes": [4096, 8192], "reuseAllowed": true } }
```

Also ran `go build ./...`, `go vet ./...`, and the full `pkg/policy` unit test suite locally (all passing). `pkg/venafi/tpp`'s own test suite requires live TPP test credentials not set up for this exact package's test harness, so it wasn't run directly here — the two live verifications above exercise the same code paths against a real server.


simeoncybr (approved):

pkg/venafi/tpp/tpp.go
```diff
@@ -981,7 +1024,13 @@ func (sp serverPolicy) toPolicy() (p endpoint.Policy) {
 	} else {
 		p.UpnSanRegExs = []string{}
 	}
-	if sp.KeyPair.KeyAlgorithm.Locked {
+	if sp.KeyPair.PkixParameterSet.Locked && len(sp.KeyPair.PkixParameterSet.Values) > 0 {
+		configs, err := allowedKeyConfigurationsFromPkixParameterSet(sp.KeyPair.PkixParameterSet.Values)
+		if err != nil {
+			panic(err)
```
wallrj: This panic fires on server-controlled input: any OID missing from the 8-entry table (a future PQC algorithm, Ed25519, a new RSA size) crashes the whole process. Both callers of `toPolicy` have error returns ([connector.go#L1649](https://github.com/Venafi/vcert/blob/e4a77c8783c2642a593a0de1484a9f43013ad5b3/pkg/venafi/tpp/connector.go#L1649), [#L1686](https://github.com/Venafi/vcert/blob/e4a77c8783c2642a593a0de1484a9f43013ad5b3/pkg/venafi/tpp/connector.go#L1686)) and there's no `recover()` in `pkg/`, so SDK consumers (cert-manager, for one) crash outright. `BuildPolicySpecificationForTPP` returns a clean error for the identical condition, so the same folder state gives a tidy error from `getpolicy` but a crash from `enroll`.

I know the surrounding code already panics on `KeyType.Set`, but those inputs are closed-world; this one is open-ended. Suggest giving `toPolicy` an error return rather than extending the pattern.

wallrj (commented):
Review of the PKIX parameter set handling. The inline comments cover a panic on server-controlled input, four ways the `getpolicy` → `setpolicy` round trip breaks, a fail-open fallback, and some duplication.

One more finding that's outside the diff: `toZoneConfig` ([tpp.go#L906](https://github.com/Venafi/vcert/blob/e4a77c8783c2642a593a0de1484a9f43013ad5b3/pkg/venafi/tpp/tpp.go#L906)) wasn't updated for `PkixParameterSet`, so the zone's default `KeyConfiguration` is still derived solely from the deprecated fields. On a PKIX-locked folder where the deprecated `KeyAlgorithm` is empty/unlocked, `zc.KeyConfiguration` stays nil (the `KeyType.Set("")` error is swallowed by the bare return), `UpdateCertificateRequest` falls back to RSA-2048 ([endpoint.go#L522-L548](https://github.com/Venafi/vcert/blob/e4a77c8783c2642a593a0de1484a9f43013ad5b3/pkg/endpoint/endpoint.go#L522-L548)), and the generated key is then rejected by the same zone's new `AllowedKeyConfigurations` — so enrolment that relies on zone defaults fails.

*with claude fable-5*

pkg/venafi/tpp/tpp.go
```diff
@@ -981,7 +1024,13 @@ func (sp serverPolicy) toPolicy() (p endpoint.Policy) {
 	} else {
 		p.UpnSanRegExs = []string{}
 	}
-	if sp.KeyPair.KeyAlgorithm.Locked {
+	if sp.KeyPair.PkixParameterSet.Locked && len(sp.KeyPair.PkixParameterSet.Values) > 0 {
```
wallrj: If TPP ever returns `PkixParameterSet{Locked: true, Values: []}` (an empty allow-list, or a serialisation quirk), this condition is false and we fall through to the deprecated `KeyAlgorithm.Locked` branch — which TPP 25.1+ no longer locks — and [tpp.go#L1061-L1083](https://github.com/Venafi/vcert/blob/e4a77c8783c2642a593a0de1484a9f43013ad5b3/pkg/venafi/tpp/tpp.go#L1061-L1083) then emits every RSA size plus all supported curves: an unrestricted policy, no error, no log. That's the fail-open behaviour this PR is fixing. Suggest treating `Locked && len(Values) == 0` as an error, like the unknown-OID case. Same pattern in `BuildPolicySpecificationForTPP`.

pkg/venafi/tpp/tpp.go
```diff
@@ -853,6 +862,40 @@ type serverPolicy struct {
 	WildcardsAllowed      bool
 }
 
+// allowedKeyConfigurationsFromPkixParameterSet decodes a list of PKIX parameter set OIDs, as returned by
+// TPP 25.1+'s Certificates/CheckPolicy KeyPair.PkixParameterSet.Values, into AllowedKeyConfigurations.
+// The OID table (policy.PkixToKeyAlgorithms) is shared with pkg/policy's getpolicy/setpolicy CLI path,
+// which decodes the same OIDs from the same TPP API family.
+func allowedKeyConfigurationsFromPkixParameterSet(oids []string) ([]endpoint.AllowedKeyConfiguration, error) {
```
wallrj, line 866: This duplicates the decode loop in `BuildPolicySpecificationForTPP` ([policyUtils.go#L627-L645](https://github.com/Venafi/vcert/blob/e4a77c8783c2642a593a0de1484a9f43013ad5b3/pkg/policy/policyUtils.go#L627-L645)) — same map lookup, same RSA/ECDSA switch, error strings already diverged by the `tpp: ` prefix — and `PkixToKeyAlgorithms` is a second hand-maintained inverse of `KeyAlgorithmsToPKIX`. The missed `TppRsaKeySize` update (see comment there) shows what drift across sibling tables costs. One shared decoder in `pkg/policy` returning `(rsaSizes []int, curves []string, err error)` would serve both paths, and the inverse map could be derived from the forward map at init so future OID additions are single-site. The anonymous `PkixParameterSet` struct above also re-declares the wire shape of `policy.LockedArrayAttribute` ([policyStructures.go#L202](https://github.com/Venafi/vcert/blob/e4a77c8783c2642a593a0de1484a9f43013ad5b3/pkg/policy/policyStructures.go#L202)).

pkg/venafi/tpp/tpp.go
```diff
@@ -853,6 +862,40 @@ type serverPolicy struct {
...
+	var rsaSizes []int
+	var curves []certificate.EllipticCurve
+	for _, oid := range oids {
+		entry, ok := policy.PkixToKeyAlgorithms[oid]
+		if !ok {
+			return nil, fmt.Errorf("tpp: policy allows unrecognised PKIX parameter set OID %q; vcert's OID table may need updating", oid)
```
wallrj: Nit: `unrecognised` → `unrecognized` (US spelling), here and at policyUtils.go:630, plus the test names and `strings.Contains` assertions that match these strings.

pkg/policy/policyUtils.go
```diff
@@ -31,6 +31,7 @@ var KeyAlgorithmsToPKIX = map[string]map[string]string{
 		"2048": "1.3.6.1.4.1.28783.10.1.1.2048",
 		"3072": "1.3.6.1.4.1.28783.10.1.1.3072",
 		"4096": "1.3.6.1.4.1.28783.10.1.1.4096",
+		"8192": "1.3.6.1.4.1.28783.10.1.1.8192",
```
wallrj, line 38: 8192 is added to both OID maps but not to `TppRsaKeySize` ([policyUtils.go#L19](https://github.com/Venafi/vcert/blob/e4a77c8783c2642a593a0de1484a9f43013ad5b3/pkg/policy/policyUtils.go#L19)), so even a single-value `rsaKeySizes: [8192]` spec — this PR's headline scenario — is rejected by `validateKeyPair` ([policyUtils.go#L167-L169](https://github.com/Venafi/vcert/blob/e4a77c8783c2642a593a0de1484a9f43013ad5b3/pkg/policy/policyUtils.go#L167-L169), "specified rsaKeySizes doesn't match with supported ones"). The default-key check at [L315](https://github.com/Venafi/vcert/blob/e4a77c8783c2642a593a0de1484a9f43013ad5b3/pkg/policy/policyUtils.go#L315) has the same gap.

pkg/policy/policyUtils.go
```diff
@@ -596,34 +618,60 @@ func BuildPolicySpecificationForTPP(checkPolicyResp CheckPolicyResponse) (*Polic
...
-
-	if strings.ToUpper(policy.KeyPairResponse.KeyAlgorithm.Value) == "RSA" {
-		//resolve rsaKeySizes
-		if policy.KeyPairResponse.KeySize.Value > 0 {
-			if policy.KeyPairResponse.KeySize.Locked {
-				keyPair.RsaKeySizes = []int{policy.KeyPairResponse.KeySize.Value}
-			} else {
-				shouldCreateDefKeyPair = true
-				defaultKeyPair.RsaKeySize = &policy.KeyPairResponse.KeySize.Value
+	if policy.KeyPairResponse.PkixParameterSet.Locked && len(policy.KeyPairResponse.PkixParameterSet.Value) > 0 {
```
wallrj: Two things with this branch:

1. It bypasses the `defaultKeyPair`/`shouldCreateDefKeyPair` resolution the legacy branch performs, so `default.keyPair` disappears from `getpolicy` output for TPP 25.1+ folders, and a subsequent `setpolicy` round trip erases the folder's defaults.
2. When `Locked` is false the PKIX values (recommended-only) are dropped entirely — `KeyPairResponse` has no field to carry a PKIX default.

Also, `Locked && len(Values) == 0` falls through to the deprecated (now-unlocked) fields — see the matching comment on the tpp.go branch.

pkg/policy/policyUtils.go
```diff
@@ -596,34 +618,60 @@ func BuildPolicySpecificationForTPP(checkPolicyResp CheckPolicyResponse) (*Polic
...
+		//TPP 25.1+: allowed key algorithms are locked via the PKIX parameter set OID list rather
+		//than the deprecated KeyAlgorithm/KeySize/EllipticCurve fields below.
+		var keyTypes []string
+		var rsaKeySizes []int
+		var curves []string
+		for _, oid := range policy.KeyPairResponse.PkixParameterSet.Value {
+			alg, ok := PkixToKeyAlgorithms[oid]
+			if !ok {
+				return nil, fmt.Errorf("policy allows unrecognised PKIX parameter set OID %q; vcert's OID table may need updating", oid)
```
wallrj: One unrecognised OID aborts the entire policy read (and the tpp.go equivalent panics). A folder that adds, say, ML-DSA alongside RSA-2048 makes vcert unusable for that zone even for clients only requesting RSA, until a vcert with an updated table ships. Consider skipping unknown OIDs with a warning — still fail-closed for enforcement — and erroring only when every OID is unknown.

pkg/policy/policyUtils.go
```diff
@@ -596,34 +618,60 @@ func BuildPolicySpecificationForTPP(checkPolicyResp CheckPolicyResponse) (*Polic
...
+				keyTypes = append(keyTypes, alg.KeyType)
+			}
+			switch alg.KeyType {
+			case "RSA":
+				rsaKeySizes = append(rsaKeySizes, alg.KeySize)
+			case "ECDSA":
+				curves = append(curves, alg.Curve)
 			}
 		}
+		keyPair.KeyTypes = keyTypes
```
wallrj: Round-trip bug: `keyTypes` gets `"ECDSA"` here, but the `setpolicy` write-back map `KeyAlgorithmsToPKIX` is keyed `"ECC"` ([policyUtils.go#L36](https://github.com/Venafi/vcert/blob/e4a77c8783c2642a593a0de1484a9f43013ad5b3/pkg/policy/policyUtils.go#L36)). `BuildTppPolicy` copies `KeyTypes[0]` into `KeyAlgorithm` ([policyUtils.go#L410](https://github.com/Venafi/vcert/blob/e4a77c8783c2642a593a0de1484a9f43013ad5b3/pkg/policy/policyUtils.go#L410)), and the lookup at [connector.go#L1179](https://github.com/Venafi/vcert/blob/e4a77c8783c2642a593a0de1484a9f43013ad5b3/pkg/venafi/tpp/connector.go#L1179) misses with no else branch, so no `PKIX Parameter Set` attribute is written — and on TPP ≥ 25 the legacy `Key Algorithm` branch is skipped too. Net effect: `getpolicy` → `setpolicy` of an ECDSA-locked folder silently leaves the target folder with no key-algorithm restriction at all.

pkg/policy/policyUtils.go
```diff
@@ -596,34 +618,60 @@ func BuildPolicySpecificationForTPP(checkPolicyResp CheckPolicyResponse) (*Polic
...
+		keyPair.RsaKeySizes = rsaKeySizes
```
wallrj: The multi-valued output here (e.g. `rsaKeySizes: [4096, 8192]`, this PR's own test case) is rejected by vcert's own `setpolicy` validation: `validateKeyPair` errors on more than one value ([policyUtils.go#L156-L172](https://github.com/Venafi/vcert/blob/e4a77c8783c2642a593a0de1484a9f43013ad5b3/pkg/policy/policyUtils.go#L156-L172)), and mixed RSA+ECDSA fails on `keyTypes` — yet README-CLI-PLATFORM.md documents `getpolicy --file` output as `setpolicy` input. Populating the spec's existing `pkixParameterSet` field ([policySpecification.go#L32](https://github.com/Venafi/vcert/blob/e4a77c8783c2642a593a0de1484a9f43013ad5b3/pkg/policy/policySpecification.go#L32)), which `BuildTppPolicy` already round-trips verbatim, would make the round trip lossless and also sidestep the ECC/ECDSA mismatch above.

pkg/venafi/tpp/tpp.go
wallrj-cyberark: Agreed, and fixed in c5564da. `toPolicy` now returns `(endpoint.Policy, error)`; the four `panic`s inside it (including the pre-existing `KeyType.Set` ones, since the error return made them free to convert) are now returned errors, and both callers in `connector.go` propagate them. The `panic("unreachable")` in the `domainRegex` closure is left alone — that one really is unreachable.

I also took the point that a single unknown OID shouldn't be fatal at all; see the reply on policyUtils.go:630.

pkg/venafi/tpp/tpp.go
wallrj-cyberark: Fixed in c5564da. The condition is now just `sp.KeyPair.PkixParameterSet.Locked`, and the shared decoder errors on an empty list, so `Locked && len(Values) == 0` fails closed instead of falling through to the deprecated (and, on 25.1+, unlocked) fields. `BuildPolicySpecificationForTPP` gets the same treatment via the same decoder.

Guarded by `TestConvertServerPolicyToInternalPolicy_PkixParameterSet/locked_with_an_empty_list_is_an_error,_not_an_unrestricted_policy` and `TestBuildPolicySpecificationForTPPPkixParameterSetLockedEmpty`.

pkg/venafi/tpp/tpp.go
wallrj-cyberark: All three points fixed in c5564da:

- The decode loop now lives once, in `policy.DecodePkixParameterSet`, returning a `DecodedPkixParameterSet` with the recognised OIDs plus the decoded key types, RSA sizes and curves. `allowedKeyConfigurationsFromPkixParameterSet` is reduced to mapping that onto `AllowedKeyConfiguration`, and `BuildPolicySpecificationForTPP` onto the policy specification. One error-string source, so no more `tpp: ` divergence beyond a single wrapping prefix.
- `PkixToKeyAlgorithms` is now derived from `KeyAlgorithmsToPKIX` at initialisation rather than hand-maintained, so adding an OID is a single-site change. `TestPkixToKeyAlgorithmsIsInverseOfKeyAlgorithmsToPKIX` asserts the round trip for every entry — and, since the `TppRsaKeySize` gap you found below is exactly this class of drift, it also asserts every RSA size in the table is accepted by validation.
- The anonymous struct is replaced by `policy.LockedArrayAttribute`.

pkg/venafi/tpp/tpp.go
wallrj-cyberark: Fixed in c5564da — the affected strings were rewritten by the shared-decoder change anyway, and both they and the test names/assertions now use "unrecognized"/"recognizes".

pkg/policy/policyUtils.go
wallrj-cyberark: Good catch, and it is exactly the drift the derived inverse map is meant to prevent. Fixed in c5564da: `TppRsaKeySize` gains 8192, which covers both `validateKeyPair` and the default-key check (that one reads the same slice). 8192 is a real supported size elsewhere in vcert — `certificate.AllSupportedKeySizes()` has included it for a while, and README-PLAYBOOK.md documents it — so the table was the outlier.

`TestPkixToKeyAlgorithmsIsInverseOfKeyAlgorithmsToPKIX` now fails if any RSA size gets a PKIX OID without also being added to `TppRsaKeySize`.

pkg/policy/policyUtils.go
wallrj-cyberark: Both fixed in c5564da:

1. The unlocked case now populates `defaults.keyPair`: `pkixParameterSetDefault` from the first OID (TPP takes a single default), plus the decoded `keyType`/`rsaKeySize`/`ellipticCurve` for readability. `BuildTppPolicy` already round-trips `PkixParameterSetDefault`, so a `getpolicy` → `setpolicy` of a recommend-only folder now preserves the default instead of erasing it. Covered by `TestBuildPolicySpecificationForTPPPkixParameterSetNotLocked`.
2. `Locked && len == 0` is now an error — see the reply on the tpp.go branch.

One thing I have deliberately not done: in the *locked* case no `defaults.keyPair` is emitted. That matches the legacy branch, which also only creates defaults when a field is unlocked, and `CheckPolicy`'s `KeyPair` response has no "PKIX Parameter Set Default" field for us to read even if we wanted to.

pkg/policy/policyUtils.go
wallrj-cyberark: Agreed — fail-closed shouldn't mean fail-shut. Fixed in c5564da: `DecodePkixParameterSet` skips OIDs it doesn't recognise with a `log.Printf` warning and errors only when *nothing* in the list is recognised (or the list is empty). Skipping narrows the algorithms vcert will permit relative to the server, so it is still fail-closed, and a folder that adds ML-DSA alongside RSA-2048 stays usable for RSA clients.

Covered by `.../an_unrecognized_OID_alongside_a_recognized_one_is_ignored` and `TestBuildPolicySpecificationForTPPPkixParameterSetPartiallyUnrecognized`, with the all-unknown case still erroring.

pkg/policy/policyUtils.go
wallrj-cyberark: Fixed in c5564da, at the root: `PkixToKeyAlgorithms` is now derived from `KeyAlgorithmsToPKIX`, so the decoded key type is whatever the forward map is keyed by — `"ECC"`, not `"ECDSA"`. The `KeyAlgorithmsToPKIX` lookup at connector.go:1179 therefore hits, and it cannot drift again without `TestPkixToKeyAlgorithmsIsInverseOfKeyAlgorithmsToPKIX` failing.

Belt and braces: the spec's `pkixParameterSet` is now populated too (see the reply below), so on TPP ≥ 25 the write-back takes the `tppPolicy.PkixParameterSet != nil` branch and never depends on the `KeyAlgorithm` lookup at all. `TestBuildPolicySpecificationForTPPPkixParameterSetEcc` asserts the emitted keyType has an entry in `KeyAlgorithmsToPKIX`.

pkg/policy/policyUtils.go
wallrj-cyberark: Adopted in c5564da. `getpolicy` now emits `pkixParameterSet` with the OIDs alongside the decoded `keyTypes`/`rsaKeySizes`/`ellipticCurves`, and `validateKeyPair` returns early — validating the OIDs instead — when `pkixParameterSet` is populated, since it is what `BuildTppPolicy` writes back and the decoded arrays are only the readable view of it. So the multi-valued output no longer trips vcert's own `setpolicy` validation, and the round trip is lossless.

`TestBuildPolicySpecificationForTPPPkixParameterSet` now feeds its own output through `validateKeyPair` to keep that honest, and README-CLI-PLATFORM.md documents the new field next to the existing "only a single value is allowed" note.

wallrj-cyberark:
The `toZoneConfig` finding was right, and it is fixed in c5564da. `toZoneConfig` now derives the zone default from `PkixParameterSet` when it is present, taking the first algorithm TPP lists (TPP returns them in its own preference order), and returns an `error` so a failure is no longer swallowed by a bare `return`. The legacy path keeps its silent "no algorithm named, so no default to offer" behaviour.

`TestToZoneConfigPkixParameterSet/the_zone_default_satisfies_the_zone_policy` builds a request from the zone default and runs it through `ValidateCertificateRequest` against the same zone's policy, which is the exact failure you described.

All ten inline findings plus this one are addressed in c5564da. Two of them changed behaviour beyond a straight fix, so flagging them explicitly:

- **Unknown OIDs are now skipped with a warning, not fatal.** Erroring only when nothing in the list is recognised. This is the one place I traded strictness for availability; the result is still a subset of the server's allow-list, so enforcement stays fail-closed.
- **`getpolicy` output gained a `pkixParameterSet` field** for TPP 25.1+ folders, and `validateKeyPair` now short-circuits when it is set. That is what makes the documented `getpolicy --file` → `setpolicy` round trip work for a multi-algorithm folder, but it does mean `keyTypes`/`rsaKeySizes`/`ellipticCurves` are informational rather than authoritative in that case.

Kept as a new commit rather than an amend, deliberately: `approver-policy-enterprise`'s `go.mod` pins this branch by pseudo-version, and the last force-push orphaned it.

Still outstanding on this PR: the `pkg/venafi/tpp` package's `init` in `connector_test.go` demands a live TPP, so the new unit tests there were run locally with that `init` short-circuited. They need a run in CI with real credentials to count as verified.

wallrj (commented):
Re-reviewed at 74dcacc. All nine inline findings and the `toZoneConfig` one from the review body are addressed — I checked the code, not just the replies:

- `toPolicy` returns errors (pre-existing `KeyType.Set` panics converted too) and both callers propagate.
- `Locked` with an empty or all-unrecognised OID list is an error, not a fall-through to the unlocked deprecated fields.
- Single shared `policy.DecodePkixParameterSet`; `PkixToKeyAlgorithms` derived from `KeyAlgorithmsToPKIX` at init; anonymous struct replaced by `policy.LockedArrayAttribute`.
- Unknown OIDs skipped with a warning, error only when nothing is recognised.
- `TppRsaKeySize` gains 8192, with the inverse-map test guarding future drift.
- Decoded key type is now `"ECC"` (derived from the forward map), and `pkixParameterSet` is emitted verbatim with `validateKeyPair` accepting it, so the `getpolicy` → `setpolicy` round trip is lossless; README updated.
- Unlocked PKIX values become `defaults.keyPair` with `pkixParameterSetDefault`; agreed that the locked case emitting no default matches the legacy branch and CheckPolicy offers nothing to read there.
- `toZoneConfig` derives the zone default from the first PKIX OID and returns errors.
- US spellings throughout the new strings and tests.

Verified locally at 74dcacc: `go build ./...` passes and the `pkg/policy` tests pass, including the new PKIX ones. The `pkg/venafi/tpp` tests dial a live TPP from `init` (connector_test.go:80) so I couldn't run the new conversion tests offline.

I can't resolve the threads from this account (no write access on this repo) — they're all ready to resolve from yours.

*with claude fable-5*

wallrj-cyberark:
Rebased onto master and squashed to a single commit (74dcacc). No content change from the reviewed state — the tree is byte-identical to the previous head, and the only thing that came in from master was the NGTS/playbook docs commit.

Re-ran the downstream proof against the live TPP 25.1+ test instance on this exact commit, and all ten subtests pass, including the three that were previously skipped because the key-algorithm policy was not being read:

```
--- PASS: TestTPPPluginEvaluate (55.93s)
    --- PASS: TestTPPPluginEvaluate/success-keysize-equals-locked-value
    --- PASS: TestTPPPluginEvaluate/success-dns-subdomain
    --- PASS: TestTPPPluginEvaluate/success-keysize-greater-than-locked-value
    --- PASS: TestTPPPluginEvaluate/wrong-key-type
    --- PASS: TestTPPPluginEvaluate/unsupported-key-size-1
    --- PASS: TestTPPPluginEvaluate/unsupported-key-size-2
    --- PASS: TestTPPPluginEvaluate/too-small-key-size
    --- PASS: TestTPPPluginEvaluate/empty-commonname-denied
    --- PASS: TestTPPPluginEvaluate/wildcard-commonname-denied
    --- PASS: TestTPPPluginEvaluate/wildcard-dnsname-denied
```

`wrong-key-type`, `unsupported-key-size-2` and `too-small-key-size` are the three that this change fixes.

@simeoncybr your approval was recorded against the pre-review commit, so it will need re-applying if you are happy with the changes made in response to the review.

pkg/policy/policyUtils.go
```diff
@@ -31,14 +35,111 @@ var KeyAlgorithmsToPKIX = map[string]map[string]string{
...
+		"8192": "1.3.6.1.4.1.28783.10.1.1.8192",
 	},
 	"ECC": {
 		"P256": "1.3.6.1.4.1.28783.10.2.1.256",
 		"P384": "1.3.6.1.4.1.28783.10.2.1.384",
 		"P521": "1.3.6.1.4.1.28783.10.2.1.521",
+		// ED25519 sits on its own arc rather than under the ECDSA one, because TPP treats it as a
+		// distinct algorithm; it is keyed under ECC here so that it decodes to a curve name that
+		// certificate.EllipticCurve.Set understands.
+		"ED25519": "1.3.6.1.4.1.28783.10.3.25519",
```
wallrj-cyberark, line 47: This is the substantive change. The OID is on TPP's `...10.3` arc rather than the ECDSA `...10.2.1` one, because TPP treats ED25519 as a distinct algorithm rather than another ECDSA curve.

It is nonetheless keyed under `ECC` here deliberately. The inverter treats any non-RSA key type as a curve, so this decodes to the curve name `ED25519`, which is exactly the string `certificate.EllipticCurve.Set` already accepts, and it makes `toPolicy` emit an ECDSA `AllowedKeyConfiguration` — which is what `endpoint.checkKey` matches an ED25519 key against. Giving ED25519 its own top-level key type would need changes in all three of those places, so this keeps the change minimal and consistent with how vcert already models ED25519.

pkg/policy/policyUtils.go
```diff
@@ -16,13 +17,16 @@ import (
...
 // CloudRsaKeySize represents the Key sizes supported by CyberArk Certificate Manager, SaaS for RSA Private Keys
 var CloudRsaKeySize = []int{1024, 2048, 3072, 4096}
 
-// TppEllipticCurves represents the curves supported by CyberArk Certificate Manager, Self-Hosted for ECDSA Private Keys
-var TppEllipticCurves = []string{"P256", "P384", "P521"}
+// TppEllipticCurves represents the curves supported by CyberArk Certificate Manager, Self-Hosted for ECDSA Private Keys.
+// ED25519 is an Edwards curve rather than an ECDSA one, but TPP groups it with the elliptic curves,
+// and vcert follows suit: certificate.AllSupportedCurves includes it and endpoint.checkKey validates
+// an ED25519 key against an ECDSA AllowedKeyConfiguration.
+var TppEllipticCurves = []string{"P256", "P384", "P521", "ED25519"}
```
wallrj-cyberark, line 29: This entry is required, not cosmetic. `validateKeyPair` checks a policy specification's `ellipticCurves` against this list, so without it a `getpolicy` of an ED25519-locked folder would emit `ellipticCurves: ["ED25519"]` that the `setpolicy` write-back then rejects.

This is the same failure mode as the missing RSA 8192 entry in `TppRsaKeySize` that this PR already fixes.

pkg/policy/policySpecification_test.go
```diff
@@ -259,6 +260,218 @@ func TestBuildPolicySpecificationForTPPLocked(t *testing.T) {
...
+				}
+				if !existIntInArray([]int{algorithm.KeySize}, TppRsaKeySize) {
+					t.Fatalf("RSA size %d has a PKIX OID but is missing from TppRsaKeySize, so setpolicy would reject it", algorithm.KeySize)
+				}
+			} else {
+				if algorithm.Curve != parameter {
+					t.Fatalf("OID %s maps back to curve %q, expected %q", oid, algorithm.Curve, parameter)
+				}
+				if !existStringInArray([]string{algorithm.Curve}, TppEllipticCurves) {
+					t.Fatalf("curve %s has a PKIX OID but is missing from TppEllipticCurves, so setpolicy would reject it", algorithm.Curve)
```
wallrj-cyberark, line 468: This assertion is the reason the `TppEllipticCurves` omission above cannot recur. The test already asserted every RSA size in the table is accepted by validation, which is how the missing RSA 8192 entry was found, but it had no equivalent assertion for curves — so adding a curve OID without its validation entry would have passed silently.

I verified the guard actually fires rather than assuming it: reverting just the `TppEllipticCurves` change makes this fail with `curve ED25519 has a PKIX OID but is missing from TppEllipticCurves, so setpolicy would reject it`, and it passes once restored.

pkg/policy/policySpecification_test.go
```diff
@@ -259,6 +260,218 @@ func TestBuildPolicySpecificationForTPPLocked(t *testing.T) {
...
+	}
+	if err := validateKeyPair(ps); err != nil {
+		t.Fatalf("getpolicy output is rejected as setpolicy input: %v", err)
+	}
+}
+
+// TestBuildPolicySpecificationForTPPPkixParameterSetEd25519 covers the ED25519 parameter set, which
+// TPP places on its own arc rather than under the ECDSA one. It is reported as an ECC curve so that
+// the getpolicy output round trips through setpolicy.
+func TestBuildPolicySpecificationForTPPPkixParameterSetEd25519(t *testing.T) {
```
wallrj-cyberark, line 336: Mirrors the existing `...Ecc` test. The `validateKeyPair` call at the end is the round-trip assertion that matters: it proves the `getpolicy` output for an ED25519-locked folder is valid `setpolicy` input.

pkg/venafi/tpp/tpp_test.go
```diff
@@ -431,3 +447,236 @@ func TestConvertServerPolicyToInternalPolicy(t *testing.T) {
...
+		k := p.AllowedKeyConfigurations[0]
+		if k.KeyType != certificate.KeyTypeECDSA {
+			t.Fatalf("expected ECDSA, got %v", k.KeyType)
+		}
+		if len(k.KeyCurves) != 1 || k.KeyCurves[0] != certificate.EllipticCurveP256 {
+			t.Fatalf("expected [P256], got %v", k.KeyCurves)
+		}
+	})
+
+	t.Run("a single ED25519 curve is allowed", func(t *testing.T) {
```
wallrj-cyberark, line 519: This covers the enforcement path rather than the CLI one, asserting that an ED25519 parameter set becomes an ECDSA `AllowedKeyConfiguration` carrying `EllipticCurveED25519` — the shape `checkKey` requires to permit an ED25519 request.

In fairness this subtest is unverified at runtime. This package's `TestMain` requires live TPP credentials and aborts with `TPP URL cannot be empty` before any test runs, identically on an unmodified checkout. It compiles, but please run it with credentials if you have them.

wallrj-cyberark (commented):
@simeoncybr thank you for offering to review. Since you last looked I have pushed one further change, so here is a self-review of what is new and why.

**What changed:** the PKIX table now also carries the ED25519 parameter set, `1.3.6.1.4.1.28783.10.3.25519`. Everything else in the PR is as you saw it.

**Why it belongs here.** Without it, this PR is still fail-open for one algorithm. A folder locked to ED25519 alone decodes to nothing recognisable, and a folder offering ED25519 alongside RSA or ECDSA silently drops it — precisely the reporting this PR exists to remove. TPP does support ED25519 for certificates: `AsymmetricKeyType.ED25519` is a first-class key type, the PKCS#10 parser handles ED25519 CSRs, the `X509 Certificate` class carries `PKIX Parameter Set Policy`, and the 25.1 migration maps the legacy `Key Algorithm = ED25519` value onto this same OID.

**One thing worth your judgement** is the decision to key ED25519 under `ECC` rather than give it its own top-level key type. My reasoning is in the inline comments; the short version is that it keeps the decoder, `EllipticCurve.Set` and `checkKey` consistent with how vcert already treats ED25519 everywhere else. If the vcert team would rather model it as a distinct key type, that is a larger change touching `checkKey` and the policy specification schema, and I am happy to do it that way instead.

**Please note this PR does not make ED25519 requests succeed.** It only makes vcert read and report an ED25519 policy correctly. There is a separate one-character defect in `pkg/endpoint/endpoint.go` which rejects every ED25519 CSR with `invalid key in csr`: it asserts the parsed public key to `*ed25519.PublicKey`, but Go's `x509.ParseCertificateRequest` returns `ed25519.PublicKey` by value, so the assertion can never succeed. I have deliberately kept that out of this PR to avoid mixing two unrelated fixes, and will raise it separately.

**Testing.** `go build ./...`, `go vet` and the full `pkg/policy` suite pass, including the two new assertions. I could not run `pkg/venafi/tpp`'s suite: its `TestMain` requires live TPP credentials, and it fails the same way on an unmodified checkout. The new subtest there compiles but has not been executed, so it would be worth someone running it with credentials before merge.

simeoncybr (approved):

ndevarapalli-panw (approved):

wallrj-cyberark:
This introduced a regression on TPP 25.1+, now fixed in #687.

`BuildPolicySpecificationForTPP` reads `PkixParameterSet.Locked` as "an administrator restricted this folder", and in that branch never sets `default.keyPair`. TPP 25.1+ reports `Locked: true` on every folder, so the default is always lost and callers dereferencing it segfault. `TestSetDefaultPolicyValuesAndValidate` catches it.

PR #687, opened 2026-09-03
fix: read the folder's own default PKIX parameter set (TPP 25.1+)

wallrj-cyberark:
Follow-up to #682, which this fixes a regression in.

## What is broken

On any TPP 25.1 or newer server, `getpolicy` returns a policy specification with no
`default.keyPair` at all. Anything that reads the default key algorithm gets a nil pointer.

Venafi's internal TPP test suite catches this as a segfault:

```
--- FAIL: TestSetDefaultPolicyValuesAndValidate (0.20s)
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x78f896]
	pkg/venafi/tpp/connector_test.go:2792
```

That line is `*(remoteDefault.KeyPair.KeyType)`, and `KeyType` is nil.

## Why

`BuildPolicySpecificationForTPP` treats `KeyPair.PkixParameterSet.Locked` as "an administrator has
restricted this folder". In that branch it fills in `policy.keyPair` and returns, never setting
`default.keyPair`.

But TPP 25.1+ reports `Locked: true` on **every** folder, including ones with no restriction at
all. TPP is stating the effective allowed set, not that somebody locked it down. So the "restricted"
branch is now taken always, and the default is always lost.

The default was never in that field to begin with. `CheckPolicy` returns a **separate** attribute,
`DefaultPkixParameterSet`, holding the single OID the folder recommends. vcert did not read it
anywhere. On an unrestricted folder a 25.x server returns both, in the same response:

```json
"PkixParameterSet":        { "Locked": true,  "Value": ["...10.1.1.2048", "...10.1.1.4096", ...] },
"DefaultPkixParameterSet": { "Locked": false, "Value": "1.3.6.1.4.1.28783.10.1.1.2048" }
```

## What this changes

- Add `DefaultPkixParameterSet` to the `CheckPolicy` response structs, in both places that decode
  them.
- Resolve the restriction and the default independently. `PkixParameterSet` still only becomes a
  policy restriction when it is locked; the default now comes from `DefaultPkixParameterSet`.
- Fall back to the deprecated unlocked `KeyAlgorithm`/`KeySize`/`EllipticCurve` when the folder
  names no default, or names an OID this build cannot decode. Newer TPP releases keep adding OIDs,
  so an unknown default must not lose the default entirely.
- Fix the same bug, in a quieter form, in `serverPolicy.toZoneConfig`. It picked `decoded.Oids[0]` —
  the first *allowed* algorithm — as the zone default. That is only correct when the folder's
  nominated default happens to be listed first.
- Give the `serverPolicy.KeyPair` anonymous struct a name, `serverPolicyKeyPair`. Three test
  literals were restating the whole type inline, so adding a field to it broke compilation.

## Testing

`TestSetDefaultPolicyValuesAndValidate` panicked against a TPP 25.x server before this change and
passes after it, on the same server and the same policy folder.

<details>
<summary>Before and after, decoding a real CheckPolicy response, plus the new unit tests</summary>

Before:

```
Default.KeyPair.KeyType    = <nil>
Default.KeyPair.RsaKeySize = <nil>
```

After:

```
Policy.KeyPair.RsaKeySizes = [4096 8192]   <- restriction still applied
Default.KeyPair.KeyType    = RSA           <- default restored
Default.KeyPair.RsaKeySize = 4096
```

New unit tests:

- `TestBuildPolicySpecificationForTPPPkixParameterSetDefault` — a locked set plus a nominated
  default yields both the restriction and the default.
- `TestBuildPolicySpecificationForTPPPkixParameterSetUnknownDefault` — an OID this build does not
  recognise falls back to the deprecated fields rather than producing a nil default.
- `TestToZoneConfigPkixParameterSet/the folder's own default wins over the first allowed algorithm`
- `TestToZoneConfigPkixParameterSet/an unrecognized default falls back to the first allowed algorithm`

The existing `TestBuildPolicySpecificationForTPPPkixParameterSetNotLocked` still passes, so an
unlocked parameter set is still not turned into a policy restriction.

One live test, `TestSetPolicy`, fails on the server used here for an unrelated reason: it hardcodes
a contact identity that does not exist on that instance. It is unaffected by this change.

</details>

## The OIDs this decoder skips

The `1.3.6.1.4.1.28783.10` arc is much larger than the three sub-arcs vcert decodes
(`.10.1.1.<bits>` RSA, `.10.2.1.<curve>` EC-DSA NIST, `.10.3.25519` Ed25519). A stock 25.x folder
offers around fifty parameter sets, so `getpolicy` logs one warning line listing the rest. They fall
into these families:

| Arc | Family |
| --- | --- |
| `.10.1.2.<bits>` | RSA PSS/OAEP (RFC 4055) |
| `.10.2.2.N` | EC-DSA Brainpool curves |
| `.10.2.3.N` | EC-DSA SECG curves, e.g. secp256k1 |
| `.10.3.44448` | Ed448 |
| `.10.4.N` | ML-DSA |
| `.10.5.1.N`, `.10.5.2.N` | SLH-DSA, pure-hash and pre-hash |
| `.10.6.N` | FN-DSA |
| `.10.7.N` | Composite ML-DSA with a classical algorithm |
| `.10.8.N` | DSA, obsolete |

vcert has no `certificate.KeyType` for any of them, so skipping is the only correct behaviour and
should stay that way.

It is tempting to fold the RSA PSS/OAEP sets into plain RSA at the same size, since the modulus
sizes match exactly. That would be a fail-open bug. An RSA-PSS key carries a different
SubjectPublicKeyInfo algorithm identifier (`1.2.840.113549.1.1.10`) to rsaEncryption
(`1.2.840.113549.1.1.1`), so reporting a PSS-only folder as permitting RSA would let a caller
approve a key TPP then rejects — the exact class of bug #682 set out to remove.


simeoncybr (approved):

ndevarapalli-panw (approved):

wallrj-cyberark:
@simeoncybr @ndevarapalli-panw please re-review before merging. Your approvals are against `3ce12a1`, and that commit has four defects. The new commit `d0fb431` fixes them.

GitHub has kept both approvals rather than dismissing them, so this PR can currently be merged on a review of code we now know is wrong. Apologies for the churn — reviewing the second commit on its own is enough.

## What was wrong

I went back and tested the change harder, including the `getpolicy` → `setpolicy` round trip. Four problems, all introduced by the first commit:

1. **The default was not checked against the allowed set.** `DefaultPkixParameterSet` was used as the zone default without confirming the locked `PkixParameterSet` contains it. A folder whose default is inherited from a parent, or was set before an administrator narrowed the allowed list, produced a zone default that the same folder's policy then rejects. Every enrolment relying on the zone default would fail its key check.

2. **`getpolicy` output stopped round-tripping through `setpolicy`.** `validateDefaultKeyPairWithPolicySubject` compared the default only against the first entry of each allowed list. Once a locked set can name several algorithms, a default that is not listed first is rejected. My own subtest "the folder's own default wins over the first allowed algorithm" produced a policy file that `vcert setpolicy` refuses to read back.

3. **The deprecated fields leaked into 25.1+ folders.** The `KeyAlgorithm`/`KeySize`/`EllipticCurve` fallback ran whenever the default OID was unrecognised, even with a locked `PkixParameterSet` present. On such folders those fields are stale and unlocked and need not agree with the locked set.

4. **A locked set with an unrecognised default gave no default at all.** The fallback to the first allowed algorithm was gated on the set being unlocked, so `getpolicy` emitted no `default.keyPair` while `toZoneConfig` returned one for the same folder. That is the nil default this PR set out to remove, still reachable.

<details>

<summary>The four tests, run against the first commit — each fails without the fix and passes with it</summary>

```
--- FAIL: TestBuildPolicySpecificationForTPPPkixParameterSetDefault
    getpolicy output is rejected as setpolicy input:
    policy default rsaKeySize doesn't match with policy's rsaKeySize value

--- FAIL: TestBuildPolicySpecificationForTPPPkixParameterSetDefaultNotAllowed
    expected the default to fall back to the first allowed size 4096, got 2048

--- FAIL: TestBuildPolicySpecificationForTPPPkixParameterSetUnknownDefault
    expected default RSA 2048, got RSA 3072

--- FAIL: TestValidateDefaultKeyPairWithPolicySubjectMembership
    a default that is in the allowed list was rejected:
    policy default keyType doesn't match with policy's keyType value
```

</details>

## What `d0fb431` changes

`DecodedPkixParameterSet` gains a `Default` method that both `BuildPolicySpecificationForTPP` and `toZoneConfig` call. It returns the nominated default when the allowed set contains it, and the first allowed algorithm otherwise, with one warning when the nomination is unrecognised or not allowed. Both call sites previously had their own copy of the rule and only one of them warned.

`validateDefaultKeyPairWithPolicySubject` now checks membership rather than the first entry. This is a no-op away from the PKIX path: `validateKeyPair` caps `keyTypes`, `rsaKeySizes` and `ellipticCurves` at one entry unless `PkixParameterSet` is set, so for every pre-25.1 specification membership and equality are the same test.

Two things worth noting for the review:

- **The pre-25.1 branch is now byte-for-byte identical to `master`.** The whole net diff is confined to the PKIX branch, the new helper, the membership check and the struct reuse.
- **`serverPolicy` now uses `policy.KeyPairResponse` directly.** The type I added in the first commit was field-for-field the same and decoded the same `CheckPolicy` JSON. Reusing it stops the next TPP attribute being added to one struct and silently missed by the other.

## Two things I have deliberately left out

**The `KeySizes[0]` clamp.** `UpdateCertificateRequest` treats `zc.KeyConfiguration.KeySizes[0]` as a minimum and bumps any smaller explicit `-key-size` up to it. Changing which OID supplies the zone default therefore changes which explicit key sizes survive. This misfires on `master` too, whenever the first allowed size is not the smallest, so it is not new here. Fixing it means touching `pkg/endpoint`, which is shared with the other connectors, so I would rather raise it separately than widen this PR. Say the word if you would prefer it in scope.

**A test for an unlocked `PkixParameterSet` combined with a `DefaultPkixParameterSet`.** `master` ignored the nomination in that case and always took the first allowed OID; this PR honours it. I think that is right and consistent with the locked case, but nothing covers it yet. Happy to add the subtest if you would like it before merge.

For what it is worth, on the 25.x server I tested against, every folder nominates a default that is inside its own allowed set and listed first. So finding 1 above is a guard rather than something I reproduced live.

